PR #11, opened 2015-08-06
Trans2_ch_bus_ar

Mechtilde:
I finished the German translation of chapter 12. Please push it to maint and to master. Then we can read it in the nightly builds


gjanssens:
Done. Thanks for your work.

PR #12, opened 2015-09-13
Trans ch invest

Mechtilde:
This is the first part of the German translation for Chapter Investments, I activate the entry in gnucash-guide.xml, too.


jralls:
Don't you want this on the maint branch so that it will get released in two weeks rather than in two years?

Mechtilde:
hello John,

yes I want it also in maint to be quick visible.

Mechtilde

jralls:
I patched it into maint and subsequently merged it to master. Please checkout maint and use that as the base for future feature branches so that 'pull' will work correctly and there's no confusion.

I also had to make a correction: adjacent hyphens (--) aren't permitted in comments and you'd commented out a section that included a command line with arguments. Please run either <tt>make html</tt> or <tt>xmllint</tt> to check your work before committing. Since I can't read German well enough to check your work it's advisable to do the former and proof-read the results in a browser.

PR #13, opened 2015-09-14
Corr ch oview

Mechtilde:
added new parts from the English text and translate it


jralls:
Patched into maint and subsequently merged to master. Please checkout the maint branch and use that as the basis for future feature branches.

PR #16, opened 2015-09-20
Corr ch basics

Mechtilde:
hope this pull request has a better quality than the last one.

I ran xmllint and build the mobi version successfully.

I rebased it with maint


jralls:
Thanks, that worked quite well. There are still link errors for the segments that haven't been translated, but that doesn't seem to affect anything besides those links.

PR #19, opened 2015-09-23
Corr ch txns

Mechtilde:
based on maint
build mobi successful


jralls:
Merged, after deleting the stray old commits at the base of the PR.

PR #22, opened 2015-09-30
Trans ch cc

Mechtilde:
rebase to maint built mobi successfully


jralls:
Looks like you rebased a branch based on master onto maint. Try this to clean up:

    git checkout maint
    git checkout -b trans-ch-cc-2
    git cherry-pick master..trans-ch-cc #note 2 dots

Then look at the history of the result and make sure that it has exactly what you want. If it does close this and submit a new PR from trans-ch-cc-2. If it doesn't, delete trans-ch-cc-2 and try again, this time cherry-picking each commit, then check history again.

PR #25, opened 2015-10-03
Trans ch loans

Mechtilde:
based on maint 
built mobi successfully
added ch_cc.xml to de/Makefile.am


jralls:
Thanks.

This is the second PR with a minor typo that I fixed. Since you say that you built it successfully I wonder if you're fixing the same thing and then forgetting to commit it?

PR #28, opened 2015-10-11
update screenshots for PDF of ch_basics

Mechtilde:
this branch based on maint 
contains only better formatted screenshots


jralls:
Merged. Thanks!

PR #29, opened 2015-10-11
update screenshots for better PDF of ch_txns

Mechtilde:
this branch based on maint too
contains only better formatted screenshots
for this time these are all updates of screenshots


jralls:
Thanks!

PR #30, opened 2015-10-11
Update ch oview

Mechtilde:
update chapter oview to latest English version
rebased it to latest maint some minutes ago


Mechtilde:
I comment out some lines in guide/de/index.xml

I hope now it works
I don't know whether I understand that file correct.

jralls:
There were a few left, two in index.xml and one in ch_oview.xml which I commented out.

Index.xml makes guide/de/gnucash-guide/ix01.html, which is rapidly becoming obsolete. Since neither of the other languages have one maybe you should just remove it.

PR #31, opened 2015-10-22
Images ch basics

Mechtilde:
I have removed guide/de/index.xml and resized 2 pictures in chapter Basics. I rebased it to maint. So it should work


jralls:
Thanks. Merged with some fixes.

PR #35, opened 2015-11-02
Corr ch invest

Mechtilde:
add some more translation of chapter invest an German images

rebased to maint and built pdf and mobi successfully


jralls:
Thanks! Just one more section to go...

PR #36, opened 2015-11-05
change paper format to A4 for German PDF

Mechtilde:
now it is possible to create the German PDf with paper format A4.

I looked in to the Japanese Makefile.am and found there this solution.

Can you add it to all relevant versions?


jralls:
Wow, that was easy. I was put off of the Japanese version because it has paper size info in its fop.xconf, and none of the other directories had that.
Merged, thanks, after applying it to the other three European Makefile.am instances.

PR #37, opened 2015-11-06
Lektorat

Mechtilde:
proofread using spell checker

adding 2 images resized to paper size


jralls:
Thanks, but please remember to run make check.

PR #39, opened 2015-12-13
Trans ch capgain

Mechtilde:
add German translation of chapter capgain
add me to the list of translators

build pdf and mobi successfully
xmllint without any problems


jralls:
Thanks!

PR #42, opened 2015-12-30
Trans bus setup

Mechtilde:
add some more translation
add the remaining chapters
also in Makefile.am and gnucash-guide.xml

rebased to maint
build mobi and pdf sucessfully


jralls:
Thanks.

PR #45, opened 2016-01-02
Screenshots

Mechtilde:
These are the last screenshots and corrections for now in this rate.

run make check &&make pdf mobi successfully


jralls:
Thanks!

PR #46, opened 2016-01-19
Help Manual translation complete

pmralbuquerque:


jralls:
Merged with a few changes:
We generally do documentation work on ```maint``` instead of ```master```, so rebased to ```maint```.
We don't have a pt_BR translation so there's no reason to specify this as pt_PT. Changed to just pt.
Commented out a couple of references that pointed to an as-yet untranslated and so commented-out id.
Added the build system changes needed to actually include the translation in the docs.

Thanks very much for your work!

PR #47, opened 2016-02-21
Added reports files

pmralbuquerque:
*Changed the guide structure moving reports to an independent section, as they affect both personal and business finances.
+Subsequently added some more files as a suggestion.


jralls:
The translation doesn't pass make check.
Please do the following:
1. Move it to the maint branch. Master is for the next major (2.8) release and won't be published until the end of 2017:
```
git branch pt-trans
git reset --hard origin/master
git checkout maint
git rebase maint pt-trans
```
2. Change the directory from pt_PT to pt. (```git mv guide/pt_PT guide/pt```).
3. Add 'pt' (no quotes) to the ```SUBDIRS``` line in ```guide/Makefile.am```.
4. Add the line ```guide/pt/Makefile``` at line 205 in ```configure.ac```.

Now you can test your changes:
```
mkdir build
cd build
../autogen.sh
../configure
make check
```
Work through fixing all of the errors; you shouldn't need to run autogen.sh or configure again, just make check. When you have it so that it passes make check, push the pt-trans branch to your Github repository, change to it with the branch drop-down, and make a new pull request.

I don't feel strongly either way about breaking up the reports chapter into multiple files. That's not required by DocBook, but if it's easier to maintain that way it's fine.

Your version of reports is about 10x larger than the C version, so you must have written a lot of new material. That's great! Take credit for it in the document headers by declaring yourself the author rather than the translator. Your command of English seems pretty solid, can you provide the expanded material in English as well?

PR #49, opened 2016-02-23
Maint

sunfish62:


gjanssens:
David, why did you close and reopen this pr ?

gjanssens:
Or even, why was this PR created in the first place ?

sunfish62:
Geert, I plead ignorance. I was trying to get my local copy updated with the latest versions, and thought that I would accomplish that by pushing or pulling something.  When that didn't work, I tried again to see whether I'd missed something. 

When that didn't work either, I nuked my local copy and restarted fresh. I expect you'll be wondering what crazy things I'm up to. It's nice to know that people are watching who actually know what this git stuff is supposed to look like.

gjanssens:
That's ok. I just couldn't figure out what you wanted us to do with this pull request. That's clearer now.

Next time you wish to update your local copy with the newest upstream commits feel free to ask me about it on gnucash-devel or on IRC (when I'm around). I'm more than happy to guide you through the process. It would be an opportunity as well to check if our development process documentation is missing some relevant parts.

fellen:
There is currently no gjanssens on IRC. :-(

David, can you explain us, which parts of http://wiki.gnucash.org/wiki/Git#GitHub are not clear enough?

Geert, I think e.g. "How do I update my github fork" is missing.

sunfish62:
Frank, 

I will simply fill you in by saying that I am officially a Git-iot (that’s “Idiot” mixed with Git). Geert already knows this, I believe, which is why he raised his question so diplomatically. 

For your edification, I point you to my last foray into Git, eloquently documented in http://lists.gnucash.org/pipermail/gnucash-devel/2015-February/038507.html <http://lists.gnucash.org/pipermail/gnucash-devel/2015-February/038507.html> and following. If you read that, you’ll see what you’re up against.

I understand on an extremely rudimentary level how version control systems work, and over my years using GnuCash, I have evolved from Cranky List Subscriber, to Cranky Bug Creator, to Less Cranky Bug Commenter, to Commenter with Rewrite Suggestions, to Commenter Who Might Stumble Onto an Actual Docs Patch Once in A While.

As for the page you reference, I suspect that the information is there, but it’s WAY over my head. This may not be not such an issue for contributors on the main project, since those folks tend to be Actual Programmers. However, the docs side of things is another matter. 

For example, the wiki describes how to set up a Repository, but doesn’t explain what that does in Gitland. It then explains creating patches, but doesn’t explain that either. Then, we get into Pull Rrequests, which again isn’t explained. This last concept is especially challenging for me, BTW. Finally, in all these sections, the directions list command-line commands, which are fine if you’re a Linux geek, but a challenge if you’re a Mac or Windows user, and even more confusing if you’re trying to use some sort of GUI interface to these tools (like SourceTree). I wouldn’t even know which commands I’ve invoked.

To be honest, the versioning system is a significant barrier to collaboration for groups of people who might have skills with writing and editing, but not with programming.

PR #51, opened 2016-02-24
Simple commit to address issue I raised on Bugzilla, Bug 762088.

sunfish62:
Correct me if this is the wrong way...

### DIFF
--- a/guide/C/ch_capgain.xml
+++ b/guide/C/ch_capgain.xml
@@ -471,36 +471,6 @@
       </orderedlist>
     </sect2>
 
-    <sect2 id="capgain_examplecaution2">
-      <title>Caution about Valuation</title>
-
-      <para>As we see in this example, for non-financial assets, it may be
-      difficult to correctly estimate the <quote>true</quote> value of an asset. It is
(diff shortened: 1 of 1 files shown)


jralls:
Thanks, merged after a rebase.

sunfish62:
John, can you explain this for me? I don't understand the significance of "merged after a rebase." 
Is there something I should have done differently to make this a smoother process? And both this request and #52 show "Closed with unmerged commits" What implications does this have on my end going forward (i.e., do I have to do something here in order to continue work?)?

jralls:
David,
You didn't do anything wrong. Once I merged your other PR I had to rebase this one so that the history would stay linear. That's what "merged after a rebase" means. That in turn meant that what I pushed onto maint was no longer the same commit as your PR because it has a different parent, so github can't tell that the PR is merged, so it reports that the PR was closed without merging the commit.

Your next step should be to pull "upstream" into the maint branch in your local (i.e. on your mac) repo and then push it back to your github repo as I explained in email yesterday. You can use that new maint as the start point for new feature branches.

PR #52, opened 2016-02-25
Patch to add note for Balance Column resize (Bug 762656)

sunfish62:


jralls:
Thanks, merged with a minor fix: We use the <guilabel> element for column headers rather than <emphasis>.

sunfish62:
Thanks John. Could you tell me again what alterations you made? The XML got eaten...

jralls:
Look at [your commit in the gnucash mirror](https://github.com/Gnucash/gnucash-docs/commit/1259a0c4006827fc8b6418a65c8d780c0b26ea14) and you'll see that I replaced ```<emphasis>...</emphasis>``` with ```<guilabel>...</guilabel>``` in two places.

sunfish62:
Ah. I tried that, but didn't notice the tag difference. Thanks.

PR #55, opened 2016-07-28
Bug 115461 Highlight need to select Account Type when adding new stk acct

goodvibes2:
Make it clearer in the Guide that you need to select Account Type before trying to select/add the Security. There are 2 commits, 2nd is just slight grammar improvement.
I had already done my commits with description of bug 115461 before bug 769256 was created, but I don't think it matters as this pull request is only to document how to work around the problem, not to actually fix it.

### DIFF
--- a/guide/C/ch_invest.xml
+++ b/guide/C/ch_invest.xml
@@ -49,7 +49,7 @@
       <varlistentry>
         <term><emphasis>Capital gains</emphasis></term>
         <listitem>
-          <para>It is the difference between
+          <para>The difference between
           the purchase and selling prices of an investment. If the selling
           price is lower than the purchase price, this is called a
           <emphasis>capital loss</emphasis>. Also known as <emphasis>realized
@@ -60,7 +60,7 @@
       <varlistentry>
         <term><emphasis>Commission</emphasis></term>
         <listitem>
-          <para>It is the fee you pay to a broker
+          <para>The fee you pay to a broker
           to buy or sell securities.</para>
         </listitem>
       </varlistentry>
@@ -68,7 +68,7 @@
       <varlistentry>
         <term><emphasis>Common stock</emphasis></term>
         <listitem>
-          <para>It is a security that
+          <para>A security that
           represents a certain fractional ownership of a company. This is what
           you buy when you <quote>buy stock</quote> in a company on the open market. This
           is also sometimes known as <emphasis>capital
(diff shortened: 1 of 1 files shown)


fellen:
Hi Chris,

committed - Thanks.

While reviewing your commits, i got a few ideas, where we could improve that text further:

Your commit:
Optional add a currency level between broker and securities, if you have different currencies at the same broker.

Terminology section:
Capital gains: explains realized, link "Realized vs. Unrealized",pending (e.g. online quote - purchase price)
Common stock: preference stocks are fractions,  too.
Shareholder: can be another company, too.
Mutual funds have a net asset value (nav), which is daily after the closing of the exchange calucated -  instead of a intraday changing prices of shares.

Pitfall in the note about moving accounts: Do not move commodity accounts e.g. from the EUR tree into the USD tree. It might confuse you later.

goodvibes2:
Hi Frank,

Thanks for reviewing and committing my mods.
I like your suggestions. I'll do them soonish...

I'm not familiar with Mutual Funds. Your definition seems to need a slight adjustment. Is this OK:

`Mutual funds have a daily net asset value (nav), which is calculated daily after the closing of the exchange - instead of intraday changing prices of shares.`

jralls:
That applies only to one flavor of Mutual Fund, open-ended ones. With an open-ended fund shares (or more correctly units) are bought and sold from the fund issuer at the previous day's NAV. There is no secondary market, at least for US-based ones.

Other mutual fund types, Closed-ended and Exchange-Traded (which are more or less the same thing, the chief difference being that generally closed-ended funds hold primary securities like stocks and bonds while exchange-traded funds hold derivatives) issue a fixed number of shares and those are traded like any other stock on an exchange.

I don't think that the difference between funds and other securities is significant for GnuCash's purposes, but somebody thought so 20 years ago and so we have the two account types.

fellen:
In the price db you can select the price type: Bid, Ask, NAV, Last, unknown.
F::Q knows: last high low net bid ask close open day_range year_range eps div cap nav price [Quote.pm:392]. But we seem to map the FQ results all to Last, at least for my commodities.

I think, at least the in GC used types should be explained somewhere, either here or in the help context of the price editor.

How about: (Open ended) mutual funds ...
or without brackets?
You are the native speakers. ;-)

jralls:
F::Q has separate fields for each of those, and for simplicity we just gets "last".  Besides, my experience with fund quotes is that F::Q is inconsistent about filling in NAV, but always has the correct number in "last".

GnuCash doesn't really have any use for price types anyway. We should just remove those labels.

fellen:
> GnuCash doesn't really have any use for price types anyway. 
GnuCash not, but some users, who hand edit their price db.

jralls:
Those users, if there actually are any, would be better served using another program for that. Particularly with the recent change to keeping a single price per day there's no way that any price except "last/close/NAV", which are the same thing after the market close, make sense to have in the pricedb.

fellen:
I would ask on the user list before I would remove something.

PR #62, opened 2016-09-16
Bug 687290: Patch to add new chapter for expense type accounts

sunfish62:


jralls:
Congratulations, your first successful PR! Thanks!

PR #63, opened 2016-09-16
Bug 769457: Add note to 4.3 of Guide

sunfish62:


jralls:
Merged after rebasing onto your previous change. Don't worry, that's normal procedure, it just means that Github doesn't recognize the merge because the change digest is different. 

Thanks!

sunfish62:
John, does your note mean that I should delete this branch, even though the Git message says "Closed with unmerged commits"?

gjanssens:
Yes.

PR #64, opened 2016-10-21
Bug-633590 Integrate Reports Descriptions into New Reports Chapter

sunfish62:
This activates a reports chapter that includes descriptions for most standard reports included with GnuCash. The descriptions are taken from wiki.gnucash.org

### DIFF
--- a/README
+++ b/README
@@ -18,6 +18,9 @@ Requirements
   scrollkeeper >=0.3.4 (its ancestor)
 * yelp (for viewing)
 
+Optional:
+* gnome-doc-utils (contains xml2po for the use of po editors like in the it translation)
+
 Additonal Requirements for Generating PDF:
 
 * Apache fop  >= 0.95
--- a/guide/C/ch_reports.xml
+++ b/guide/C/ch_reports.xml
@@ -10,193 +10,588 @@
   Translators:
                (translators put your name and email here)
 -->
- <chapter id="chapter_reports">
+ <chapter id="ch_reports">
  <title>Reports</title>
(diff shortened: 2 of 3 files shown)


jralls:
David,

Very nice. There are two minor issues, which I've fixed. 
The first is that the entity &amp;app; somehow got turned into &amp;amp;app;. Perhaps you were using an editor that needs to be told when you want to insert an entity?
The other is that while your feature branch is correctly created off of ```maint``` you submitted the PR against ```master```. That's why Frank's commit shows up in the list of commits. Not a problem, just something for you to be aware of for next time.

Because of those two changes github won't see the merge, but it's done. Thanks!

sunfish62:
John, 
Thanks again for the feedback. The amp thing was an ill-advised find and replace. 
And your point about the PR is just what I was thinking about earlier; I'd noticed that my request had extra stuff in it, but couldn't figure out what went wrong. 
Now, let me see whether I can get the next steps in this process. ..
David

 
 
  On Sat, Oct 22, 2016 at 0:24, John Ralls<notifications@github.com> wrote:   
David,

Very nice. There are two minor issues, which I've fixed. 
The first is that the entity &app; somehow got turned into &app;. Perhaps you were using an editor that needs to be told when you want to insert an entity?
The other is that while your feature branch is correctly created off of maint you submitted the PR against master. That's why Frank's commit shows up in the list of commits. Not a problem, just something for you to be aware of for next time.

Because of those two changes github won't see the merge, but it's done. Thanks!

PR #65, opened 2016-10-22
Bug 755174 Document using Lots to create investment capital gains tra…

goodvibes2:
…nsactions.
This corrects issues raised in PR#60.


guide/C/ch_reports.xml
```diff
@@ -88,6 +88,28 @@ Configuration.</para>
...
 includes extended information about commodity holdings, including information 
 about the basis, gain, and return of individual commodities.</para>
+
+  <sect4 id="rpt-advport-capgains">
+    <title>Advanced Portfolio Capital Gains</title>
+    <para>
+      The Advanced Portfolio report doesn’t use the capital gain splits to
+      calculate capital gains. It calculates the gains from the various buy and
+      sell transactions in the account without regard to whether the gains and
+      losses are recorded or not. Any realized gain splits are ignored.
```
goodvibes2: But if you go changing the capital gain split, it will not recognize it and incorrectly include it in Income.
Shall I make this clearer?

guide/C/ch_reports.xml
```diff
@@ -88,6 +88,28 @@ Configuration.</para>
...
+      account. These two splits can be in a separate transaction (as created by
+      scrubbing) or in the same transaction as the sale (this will cause 
+      incorrect future scrubbing).
+    </para>
+      
+    <para>Thus, if you wish to subclassify capital gains, say into taxable or
+      non taxable, do this as a transfer, say from Capital Gains to 
+      Capital Gains Taxable and Capital Gain Non Taxable in a separate 
+      transaction which does not include a split to the stock's account.
+    </para>
```
goodvibes2: That would be OK if it was all taxable or non-taxable.

jralls: Is the problem (both here and at 98) that the APR gets confused if there's more than one capgain split per transaction?

guide/C/ch_reports.xml
jralls: Why would changing the split make any difference? The APR looks afresh at everything every time it's run.

guide/C/ch_invest.xml
```diff
@@ -1813,6 +1813,840 @@ Income
...
+                  </row>
+                </tbody>
+              </tgroup>
+            </table>
+          </listitem>
+
+          <listitem>
+            <para>It has been reported scrubbing does not work correctly if
+              the Stock Split Assistant has been used.
+            </para>
```
jralls: I'll ask Mike to have a look at this PR. Maybe he'll have some insight into how to document it.

goodvibes2:
Hi John,
I didn't forget about the whitespace. After you said I should recommit and 'there's no real need to preserve that for posterity', I had already done multiple fixes and it would have been a pain to take the whitespace mods out, especially on the lines with multiples. I'll do it if you really want, but I don't understand the issue. The whitespace mods are really obvious in GitHub.
Maybe I should just do a general remove trailing whitespace mod on all guide & help files in a separate PR?
I'd like to get on with other documentation...

guide/C/ch_reports.xml
goodvibes2: That is a reasonable simplification.

jralls:
The point of not mixing whitespace and substantive changes in the same commit is that mixing them obscures the substantive parts of the commit. Believe it or not, Github isn't the only way we look at commits and while Github does highlight the whitespace differences other tools don't.

A general remove trailing whitespace commit is a fine idea.

goodvibes2:
Re whitespace: Fair enough - shall do a separate PR.

guide/C/ch_reports.xml
goodvibes2: Please see
https://bugzilla.gnome.org/show_bug.cgi?id=755174#c5
The APR shows gains as gains, and tries not to include gains in income.

guide/C/ch_invest.xml
```diff
@@ -1813,6 +1813,840 @@ Income
...
+                  </caption>
+                </mediaobject>
+              </screenshot>
+            </figure>
+
+            <note>
+              <para>After using the <guilabel>Scrub Account</guilabel> button only
+                the last lot is shown, so the above image is after the <guilabel>
+                Lots in Account</guilabel> window has been closed and reopened so
+                all the lots show.
```
mta1309: Are you sure this is correct?  I just tried it and after Scrub Account I saw all the lots in the lot viewer.  If what you say is correct, it sounds like a bug that should be fixed instead of documented.

guide/C/ch_reports.xml
goodvibes2: Well you are quite correct. I did a test and splitting the Orphaned-Gains-AUD into 2 splits (Income:Capital Gain (Long Term):STK and Income:Capital Gains (Short Term): STK) made no difference to the Income or anything else on the APR.
I don't understand why Mike said:

> realized gains MUST be recorded as two splits

(capitalisation of MUST is mine). I thought the reason must be the previous sentence:

> They look a lot like income and the report goes to some lengths to make sure they aren’t counted that way.

I know that was an assumption but a logical one I thought. That was why I added comment 6 which he didn't contradict...   I don't want to give Mike a hard time - I really appreciate his work :-)

guide/C/ch_reports.xml
mta1309: I don't think that including more than one capital gain in a single transaction is the issue.  The issue is that the splits must be as described in the previous two paragraphs.  I think this description is essentially correct, but it might be possible to make it less confusing.  It should be made clear that this description applies to the APR, if it is not already clear from context.  Note that each capital gain is represented by two splits which should be in the same transaction.  The second split can be in either an income or expense account.  I described it as an income account since that's what I use.

guide/C/ch_reports.xml
mta1309: I should have said "two or more splits".  I was more interested in describing what should be in the splits than how many of them there should be.

guide/C/ch_invest.xml
goodvibes2: Yes, it happened multiple times (testing on Debian). It didn't seem like a bug to me as, after you scrub, you're not really interested in the closed lots. How many lots did you scrub?

guide/C/ch_reports.xml
goodvibes2: Fixed - will be in another PR. Could you please review/commit PR's
66 Whitespace cleanup  & 67 Fix comment guide ch_invest.xml
which will save me putting whitespace back in manually.

guide/C/ch_invest.xml
```diff
@@ -1813,6 +1813,840 @@ Income
...
+        <title>Change Orphaned Gains-CCC to Gain/Loss Account</title>
+
+        <para>
+          The capital gain/loss transaction(s) created by scrubbing uses an
+          automatically created generic <emphasis>Orphaned Gains-CCC</emphasis>
+          account (where CCC is the security currency) because <application>
+          &app;</application> doesn't know which capital gain or loss account
+          should be used. After scrubbing, the user should change it to the
+          required income (or expense) gain or loss account.
+        </para>
```
goodvibes2: Thanks for picking that up. How about:
          After scrubbing, the user should edit the <emphasis>
          Orphaned Gains-CCC</emphasis> transaction split to re-assign the
          income account to a more meaningful income (or expense) gain or loss
          account (e.g. Income:Long Term Capital Gain:IBM).

guide/C/ch_reports.xml
jralls: 66 and 67 are merged.

guide/C/ch_invest.xml
```diff
@@ -1813,6 +1813,840 @@ Income
...
+
+        <para>Run a Trial Balance
+          (<menuchoice>
+            <guimenu>Reports</guimenu>
+            <guisubmenu>Income &amp; Expense</guisubmenu>
+            <guimenuitem> Trial Balance</guimenuitem>
+          </menuchoice>) and ensure, at the end of report, total debits equals
+          total credits.
+        </para>
```
goodvibes2: Done. Only Stock Split Assistant now outstanding.

goodvibes2:
All requests mods completed. Replacement is https://github.com/Gnucash/gnucash-docs/pull/69. Closed.

PR #67, opened 2016-10-27
Fix comment guide ch_invest.xml

goodvibes2:


jralls:
And two..., after rebasing on one. Thanks. That should clear the decks for the main show.

PR #71, opened 2016-11-05
Bug 755174 Help Document Lots in Account window

goodvibes2:


help/C/Help_ch_Tools_Assistants.xml
```diff
@@ -3165,4 +3165,531 @@
...
+        </para>
+
+        <para>The window proportions may be adjusted vertically by
+          dragging the vertical bar located to the right of <guilabel>Show
+          only open lots</guilabel>.
+        </para>
+
+        <para>These resizing bars can be hard to see, so move the mouse
+          pointer around until it changes to the resize pointer (<guilabel>
+          &lt;&mdash;&gt;</guilabel> in the case of the horizontal bar), then
```
goodvibes2: Did you have something in mind how to 'inline' this? I don't remember seeing anything like this in the help or guide. I see there is a docbook inlinemediaobject http://docbook.org/tdg5/en/html/inlinemediaobject.html which I could try out. I'm tempted to just say it's the horizontal or vertical resize pointer (terminology I've seen on Windows) which looks something like a double ended arrow depending on your OS and theme.
inlinemediaobject is not used in the guide or help at the moment.

help/C/Help_ch_Tools_Assistants.xml
```diff
@@ -3165,4 +3165,531 @@
...
+        </para>
+
+        <para>It is not possible to individually adjust the size of panel
+          columns.
+        </para>
+
+        <para><emphasis>Fields</emphasis></para>
+        <itemizedlist>
+          <listitem>
+            <para><guilabel>Title</guilabel> &mdash; Shows the title for the
```
jralls: So I looked a bit deeper: A bit of grepping reveals that there are 98 instances that use &amp;mdash; (all incorrectly wrapped with spaces), 157 that use ':' (4 after the &lt;/guilabel&gt; and the rest as part of the label), and 194 that use only a space. I think we should change them all to ':' as part of the label so that the colon gets the guilabel styling&mdash;not all of them as part of this PR of course, just the ones it adds.

help/C/Help_ch_Tools_Assistants.xml
jralls: If it's too hard to do the inline image then sure, yield to temptation. It's not worth much time or effort.

help/C/Help_ch_Tools_Assistants.xml
jralls: I think we're a bit out of sync. I meant that you should fix the em-dashes in what you've added for this PR in this PR. If you want to do the global fixes too, that should be a separate commit. It can still be part of this PR if you want, or a separate one. Practically, that would mean doing the changes in "tool-lots" for the ```git commit --amend``` and then whatever other changes as a second commit. 

As to the question, "Preferences &mdash; Accounting Period tab" is a figure title. I think "Preferences dialog box, Accounting Period tab" or "Accounting Period tab in the Preferences dialog box" are more correct. I also think that it would look better as a &lt;caption&gt; inside the &lt;mediaobject&gt; rather than a &lt;title&gt; on the &lt;screenshot&gt;, presuming that a caption would display under the illustration rather than above it. I don't know how that would affect the "list of figures" on the contents page.

gjanssens: Hmm, I've heard a UX expert once say "dialog box" is a developer term. Ordinary users typically don't really grasp this concept. That same woman suggested to use "window" instead. Other than that, no further comments...

goodvibes2: Thanks for clearing up what I should do re commits and PR John.
My 2 cents: I think window or dialog box is unneccessary for this. How about:
`<guilabel>Preferences, Accounting Period tab</guilabel>`?

jralls: Well, if "window" isn't necessary then why is "tab"?  Is "Preferences, Accounting Period" too terse?

goodvibes2: Good point. I think "tab" is needed, so should be "window" and "tab".  OK?
"tab" inside guilabel?

gjanssens: Agreed, try both and check for yourself what you like most.

goodvibes2:
I see we could also just use:
<guimenuitem>Preferences</guimenuitem><guilabel>Accounting Period</guilabel>
That sort of sidesteps the mac or Linux/Win issue.

help/C/Help_ch_Tools_Assistants.xml
jralls: &lt;guilabel&gt;Preferences&lt;/guilabel&gt; is under the &lt;guilabel&gt;Gnucash&lt;/guilabel&gt; menu on Macs and the &lt;guilabel&gt;Edit&lt;/guilabel&gt; menu everywhere else. I'm in favor of saying that exactly once and then just using &lt;guilabel&gt;Preferences&lt;/guilabel&gt; forever after.

goodvibes2: Just to be clear, no guimenu/guimenuitem?

jralls: I suppose to be pedantically correct it should be &lt;guimenuitem&gt; instead of &lt;guilabel&gt;, but I think both tags render the same so it doesn't really matter for our purposes.

goodvibes2: Unfortunately "title" is needed for List of Figures so I see no point in adding caption here.
Also unfortunately, the title is all bold in html, pdf & mobi (same as guilabel and guimenuitem). I'm going with:
```
        <title>The <guimenuitem>Preferences</guimenuitem> window,
          <guilabel>Accounting Period</guilabel> tab
        </title>
```

goodvibes2: Should the mdash mods for guide be a different commit than help mdash mods?

jralls: Doesn't really matter as long as emdash mods are the only changes in the commit(s).

help/C/Help_ch_Tools_Assistants.xml
goodvibes2: I found the inlinemediaobject works (html, pdf & mobi) but the images don't lineup well with the text.
I have add the 2 graphic cursor images you pointed me too as examples at the end of the paragraph.

goodvibes2:
Mods finished and force pushed to my GitHub.

jralls:
Thanks.

PR #72, opened 2016-11-07
Small addition to Help File to explain that Hidden accounts still dis…

sunfish62:
…play in Account selector. In reference to Bug 397135.


jralls:
Thanks.

PR #75, opened 2016-12-06
Bug 775592 - General Rewrite of Guide Chapter on Multiple Currencies

sunfish62:
This is a rewrite of the Guide Chapter on Multiple Currencies, to improve readability and accuracy.

### DIFF
--- a/guide/C/ch_currency.xml
+++ b/guide/C/ch_currency.xml
@@ -15,14 +15,14 @@
 <chapter id="chapter_currency">
   <title>Multiple Currencies</title>
 
-  <para>This chapter will show you how to setup your <application>&app;</application> accounts to use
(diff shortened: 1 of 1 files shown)


jralls:
Merged after rebasing onto a couple of previous commits that were needed to get the Italian help translation to build.

Note that I added a follow-on commit reformatting the &lt;literallayout&gt; blocks. Columns don't really work because they render in variable-width fonts and so the length of the strings varies depending on the font and the rendering engine. One can use tables, and I did in a couple of places, but tables can't do indents, so for the chart of accounts snippets I just pulled the currency indication next to the ends of the lines.

sunfish62:
Yes, I noted the literallayout tags (and their lack of alignment) in the original document as well, but didn't have energy to seek a different solution. Thanks for fixing those.

PR #76, opened 2016-12-06
Bug-743671: Add glossary to Tutorial and Guide.

sunfish62:


jralls:
Dave, did you write all of those definitions in your own words from personal knowledge or do you need to give credit to a source or two?

sunfish62:
I started by pulling from the wiki page, followed by looking at the German version to come up with more. I did edit most entries. A few are pulled from wikipedia or other readily available sources, again with edits. A few I pulled from the modules definitions at GnuCash.org. So, I don't really think I do, but I could, if you thought it necessary.

jralls:
Wikipedia's terms are [here](https://en.wikipedia.org/wiki/Wikipedia:FAQ/Copyright#Can_I_reuse_Wikipedia.27s_content_somewhere_else.3F). Our docs are also GFDL, so barring the unlikely case that something you borrowed is restricted, then it should be OK *with attribution*. The strictly legal thing to do would be to have a bunch of footnotes citing each article you used. That would be ugly and a bit ungainly as well as a lot of work. How about a note at the beginning of the glossary to the effect that many of the definitions are based on those found on Wikipedia and are used subject to the GFDL?

sunfish62:
I understand the desire to avoid controversy. I will note that the longest phrase taken from outside sources is 26 words long, and length of quote is a factor in fair use determination. (Most of the longer entries are actually from the wiki page).
How about: "This is a glossary of common terms used in GnuCash. Some entries here are taken from Wikipedia (en.wikipedia.org)."

sunfish62:
Oh, darn. I forgot. I pulled a couple from investopedia.com. In most cases, I did end up modifying the resulting text to better match what I thought the glossary for GnuCash should have for its user base.

jralls:
It's not just avoiding controversy, it's also the right thing to do. Fair use is pretty limited: One or two short snippets can be fair use but several dozen, even if unlrelated, is not. Note as well that fair use requires attribution. Claiming even a few words that someone else wrote as your own is plagiarism, not fair use.

I think that your blanket attribution for Wikipedia will do. Investopedia is a problem: Their [terms of use](http://www.investopedia.com/corp/termsofuse.aspx)--a contract that may extend beyond their copyright rights--prohibit republication. Period. Do you remember which terms you took from there?

sunfish62:
Agreed. 
I can rewrite the entries that came from Investopedia and use modified wikipedia language instead. Procedurally, how do I go about modifying the committed branch on my end?

jralls:
Just commit in the branch and push. Since it would be best to not perpetuate the potentially infringing material in the repository, amend the previous commit and force-push to your repo:
```
git commit -a --amend
git push -f
```

sunfish62:
Changes are now included (I think).

jralls:
Did you follow my instructions on replacing the original commit? Somehow it got turned into a merge with conflicts...

I've fixed everything up and squashed it into a single commit, with the addition of ch_reports.xml to the Makefiles.am (you need to do both C and it because of Cristian's use of a po file for translation) in a separate commit before the gnc-glossary one and pushed the result. Thanks!

sunfish62:
John, WRT replacing the original commit, I executed the commands you prescribed, without apparent error. 732f319 above was the result of those commands, and appeared to my eye to have changed things here. Clearly, that was not actually the case. I apologize that my feeble attempts at using git seem always to create more difficulties for you, and am grateful that you take the time to straighten them out for me.

One final question: github reports this request as "Closed with unmerged commits" with an option to delete the branch. I believe that I can delete the branch safely, but want to verify before doing so...

sunfish62:
John, Looking at the documentation nightly build on gnucash.org, the version included is not the most recent. There is no attribution for Wikipedia, and entries for "Accounting Period," " Balance Sheet," and "Cost Basis" do not reflect my latest edits.

jralls:
Rats, I messed up the conflict resolution when fixing it up.  I've redone it and pushed.

PR #77, opened 2016-12-06
Bug-770371 - Add note to documentation regarding Sort order options. …

sunfish62:
…Added to table in Help Section 4.3.2.3


jralls:
Merged after rebasing onto PR75. Thanks!

PR #84, opened 2017-01-28
Bug 777318 - Restructure Business Features Chapters.

sunfish62:
Please note: this commit combines 5 previous chapters (Business Intro, Business Setup, A/Receivable, A/Payable, and Payroll) into one new "Business Features" chapter. This entails removing 4 earlier chapters from the repository, which I may not get right. Please let me know whether I needed to do something different in this regard.


gjanssens:
Hi David, I'm fine with bringing all business related topics together in one chapter. I'm not sure this also means they should be put into one single file as well. For now there's not that much detail in the description of the business features yet, but if we get to a more complete documentation this may become a large body of text for one single file. Then again if we can merge them now we can also split them later if needed of course.

I guess just want to double-check you didn't do the merge because it seemed to be required. There's no technical restriction imposing that one chapter should be one file. One chapter can be spread across multiple files or the other way around, one file can hold multiple chapters as well. You can organize this as you see fit.

What motivated you to put it all in one single file ?

guide/C/Makefile.am
```diff
@@ -25,10 +25,6 @@ entities = \
 	ch_currency.xml \
 	ch_dep.xml \
 	ch_bus_intro.xml \
```
gjanssens: If we decide to keep the files merged (I personally don't have a strong opinion on this matter), I would like you to rename the remaining file as well. It now covers more than the introduction to the business features.

sunfish62: Good catch. I can change it to ch_bus_features.xml

sunfish62:
Geert, As for the merging of files, I was not aware that a chapter could span multiple files. That being said, I do feel that these elements hang together as a fundamental business feature set. As this is not an area I have a lot of experience with, I am happy retreating to separate chapters.

gjanssens:
You did quite well on the file merge as far as I can see (this is from reading the changes only - I haven't built the docs yet with your patch).

When you prefer not merging the files, you would make the same changes in the files as you now did after the file merge:

- redefine all sect2 tags as sect3
- redefine all sect1 tags as sect2
- redefine all chapter tags as sect1
- remove the links you no longer consider useful in the overview page
- optionally rename the ENTITIES in gnucash-guide.xml (instead of removing them) as they're not really chapters any more but rather sections in a chapter. For example

  `<!ENTITY chapter14 SYSTEM "ch_bus_setup.xml">`
could become 

  `<!ENTITY chapter13setup SYSTEM "ch_bus_setup.xml">`
Then use 'chapter12setup' later on when all the entities are included.

sunfish62:
Geert, the pr says you requested changes. I believe I addressed your request. Can this proceed?

sunfish62:
Geert, Is there a problem with this patch?

gjanssens:
The PR is fine. It seems I was uncertain whether you intended to revert the changes to separate chapters or not, and then got side tracked by other issues. I'll commit as soon as I get back to my workstation. Sorry for keeping you waiting.

fellen:
I saw translators getting confused. They have now different files than the english version. Next time we should consider to 
* either patch comments in their files `<!-- With commit xxxxxxx the english version of this file became [part of] ... -->`
* or explain in https://wiki.gnucash.org/wiki/Translation#Translating_the_GnuCash_Guide_and_Help how they can find them with git magics.

PR #85, opened 2017-02-06
Bug 777891 - Provide ID Tag for Glossary and reference in Guide Overview Chapter

sunfish62:
This adds an ID tag to the Guide Glossary, and adds a reference to the Glossary in the Overview Chapter of the Guide.

### DIFF
--- a/guide/C/ch_oview.xml
+++ b/guide/C/ch_oview.xml
@@ -406,6 +406,10 @@
 
     <itemizedlist>
        <listitem>
+          <para><xref linkend="gnc-gloss"></xref> - Glossary of terms used in <application>&app;</application></para>
+       </listitem>
+       
+       <listitem>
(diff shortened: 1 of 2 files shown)


gjanssens:
@sunfish62 Thanks for the PR. I chose to sqash your commits into one before applying it to the upstream repository. Your first commit can't really work without the second one so it's best to keep them in one commit.

Thanks for your work! It will appear in the next release.

sunfish62:
Of course. If I could figure out how I committed only the one change without the other, I might actually just get them both committed at once the next time. AND, if I could figure out how either to  pull back the first commit and get it right OR "squash them together" as you did, I would have done that, too.

fellen:
git commit -a 
ammends the last commit

fellen:
Sorry, -a means -all. So you have to write the full option:
git commit --amend

PR #88, opened 2017-03-27
Minor doc spelling and grammar

phil-davis:
The new and reorganised guide is good!


gjanssens:
Thank you for the corrections!

I have applied them to maint first rather than master so they will already appear in gnucash 2.6.17 instead of gnucash 2.8.

PR #89, opened 2017-04-25
Bug 781711: Add reversing transaction definition

sunfish62:


gjanssens:
Thanks!

fellen:
Just a note:
 "restores the books to the state before the error" could be read as 
"you will loose all transactions made in between".

sunfish62:
Good point. How about:
"This reversing transaction duplicates the original transaction, but with debit and credit amounts reversed. This removes the effect of the erroneous transaction from the books. After entering the reversing transaction, a corrected transaction can then be entered."

A little more wordy, but less ambiguous.

fellen:
commit eded1c2
Thanks!

PR #90, opened 2017-04-29
Bug 781714: Add Reversing Transaction text for Help document

sunfish62:
Add text to Help to explain the Add Reversing Transaction feature and its function.


gjanssens:
Thanks :)

PR #91, opened 2017-06-14
Fix typos

ajhepple:
Fix some simple typos that I spotted whilst reading the documentation recently.


gjanssens:
I have pushed your changes to our maint branch so they will appear in the next stable release already instead of only in the next major release.

Thank you for your contribution !

PR #95, opened 2017-06-30
update Makefile.am

DiMan:
final adaptation for russian translation


fellen:
Right, I will update the wiki, too.

PR #98, opened 2017-07-03
Fix 1 remaining 'incomes'

phil-davis:
Noticed this little thing. I should look just before release dates, not after.


phil-davis:
Oops - this projects wants PRs against maint.

PR #99, opened 2017-07-03
Fix 1 remaining 'incomes'

phil-davis:


gjanssens:
Thanks. It will be part of the next release. It missed 2.6.17 by a day...

phil-davis:
Yes, I saw the release notification on the mailing list and that made me go looking at diffs in the docs - by definition too late.

fellen:
Just for curiosity: are such mixed forms of singular and plural really common in English?

jralls:
There are a [few irregulars](https://www.vappingo.com/word-blog/101-words-that-are-both-plural-and-singular/). 'Income' isn't really one of them, though. It gets the plural form only if it's the object and the subject is plural.

phil-davis:
Basic summary of English: its a mess.

PR #106, opened 2018-01-20
Add Newbie Guide to the TR

christopherlam:
Not ready yet - added for discussion.
Q&A format is unsatisfatory and will be reworked.
Further use cases welcome.

### DIFF
--- a/guide/C/ch_reports.xml
+++ b/guide/C/ch_reports.xml
@@ -816,7 +816,7 @@ an empty report will be generated.</para>
           </para>
           <orderedlist>
             <listitem><para>As above</para></listitem>
-            <listitem><para><emphasis>General</emphasis>: Subtotal Table enabled</para></listitem>
+            <listitem><para><emphasis>Display</emphasis>: Subtotal Table enabled</para></listitem>
           </orderedlist>
         </answer>
       </qandaentry>
--- a/guide/C/ch_reports.xml
+++ b/guide/C/ch_reports.xml
@@ -111,7 +111,7 @@ Configuration.</para>
       <listitem>
         <para>Indenting of columns will indent the report columns for
         easier understanding of grouping and sorting. (This is now
-        enabled by default)</para>
+        enabled by default for Gnucash 3.0 onwards)</para>
       </listitem>
       <listitem>
         <para>Subtotal table is useful to understand the relationship
@@ -757,24 +757,24 @@ an empty report will be generated.</para>
       <orderedlist>
         <listitem>
           <para>Open the report options, from the
-          <emphasis>Accounts</emphasis> tab, click the Expense
-          placeholder account and <emphasis>Select
-          Children</emphasis>.</para>
+          <guilabel>Accounts</guilabel> tab, click the Expense
+          placeholder account and <guibutton>Select
+          Children</guibutton>.</para>
         </listitem>
         <listitem>
-          <para>From the <emphasis>General</emphasis> tab, Choose
+          <para>From the <guilabel>General</guilabel> tab, Choose
           relative dates “Start of Previous Year” and “End of Previous
           Year”.</para>
         </listitem>
         <listitem>
-          <para>From the <emphasis>Sorting</emphasis> tab, set
-          <emphasis>Primary Key</emphasis> to Accounts, and enable
-          <emphasis>Primary Key Subtotal</emphasis>.
+          <para>From the <guilabel>Sorting</guilabel> tab, set
+          <guilabel>Primary Key</guilabel> to Accounts, and enable
+          <guilabel>Primary Key Subtotal</guilabel>.
           </para>
         </listitem>
         <listitem>
-          <para>Set the <emphasis>Secondary Key</emphasis> to Date,
-          and set the <emphasis>Secondary Key Subtotal</emphasis> to
+          <para>Set the <guilabel>Secondary Key</guilabel> to Date,
+          and set the <guilabel>Secondary Key Subtotal</guilabel> to
           None. </para>
         </listitem>
       </orderedlist>
@@ -789,8 +789,8 @@ an empty report will be generated.</para>
       also calculate monthly sums within each account.</para>
       <orderedlist>
         <listitem><para>Select options as above</para></listitem>
-        <listitem><para>From the <emphasis>Sorting</emphasis> tab,
-        set <emphasis>Secondary Key Subtotal</emphasis> to Monthly
+        <listitem><para>From the <guilabel>Sorting</guilabel> tab,
+        set <guilabel>Secondary Key Subtotal</guilabel> to Monthly
         </para></listitem>
       </orderedlist>
     </sect3>
@@ -803,8 +803,8 @@ an empty report will be generated.</para>
       </para>
       <orderedlist>
         <listitem><para>Select options as above</para></listitem>
-        <listitem><para>From the <emphasis>Display</emphasis> tab,
-        enable the <emphasis>Subtotal Table</emphasis>.</para>
+        <listitem><para>From the <guilabel>Display</guilabel> tab,
+        enable the <guilabel>Subtotal Table</guilabel>.</para>
         </listitem>
       </orderedlist>
     </sect3>
@@ -818,9 +818,9 @@ an empty report will be generated.</para>
       The filtering text may be either a standard string, or a POSIX regular
       expression.</para>
       <orderedlist>
-        <listitem><para>From the <emphasis>Filter</emphasis> tab,
-        complete the <emphasis>Account Name Filter</emphasis>, or
-        <emphasis>Transaction Filter</emphasis>. Optionally toggle the
+        <listitem><para>From the <guilabel>Filter</guilabel> tab,
+        complete the <guilabel>Account Name Filter</guilabel>, or
+        <guilabel>Transaction Filter</guilabel>. Optionally toggle the
         regular expression check boxes to toggle full POSIX regex
         matching. Regular expressions will allow more complex
         queries e.g. the transaction filter can be set to
@@ -836,20 +836,20 @@ an empty report will be generated.</para>
       <para>This sortkey 'reconcile status' will separate reconciled,
       cleared and unreconciled transactions. This report may be useful
       in producing a printable reconciliation report.</para>
-      <para>Alternatively, the <emphasis>Reconciliation Report
-      </emphasis> will preset these defaults and requires only the
+      <para>Alternatively, the <guilabel>Reconciliation Report
+      </guilabel> will preset these defaults and requires only the
       Accounts selection.
       </para>
       <orderedlist>
-        <listitem><para>From the <emphasis>Accounts</emphasis> tab,
+        <listitem><para>From the <guilabel>Accounts</guilabel> tab,
         select the relevant bank or credit card account</para></listitem>
-        <listitem><para>From the <emphasis>General</emphasis> tab,
+        <listitem><para>From the <guilabel>General</guilabel> tab,
         choose an appropriate date range e.g. past 3 months</para></listitem>
-        <listitem><para>From the <emphasis>Sorting</emphasis> tab,
-        set the <emphasis>Primary Key</emphasis> to Reconcile
-        Status, <emphasis>Primary Key Subtotal</emphasis> to enabled,
-        and set the <emphasis>Secondary Key</emphasis> to Date. The
-        <emphasis>Secondary Date Subtotal</emphasis> is left to None
+        <listitem><para>From the <guilabel>Sorting</guilabel> tab,
+        set the <guilabel>Primary Key</guilabel> to Reconcile
+        Status, <guilabel>Primary Key Subtotal</guilabel> to enabled,
+        and set the <guilabel>Secondary Key</guilabel> to Date. The
+        <guilabel>Secondary Date Subtotal</guilabel> is left to None
         </para></listitem>
       </orderedlist>
     </sect3>
--- a/guide/C/ch_reports.xml
+++ b/guide/C/ch_reports.xml
@@ -816,7 +816,7 @@ an empty report will be generated.</para>
           </para>
           <orderedlist>
             <listitem><para>As above</para></listitem>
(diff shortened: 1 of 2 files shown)


guide/C/ch_reports.xml
```diff
@@ -774,6 +774,24 @@ an empty report will be generated.</para>
...
+      <qandaentry>
+        <question>
+          <para>Can I select tagged accounts or transactions for reporting?</para>
+        </question>
+        <answer>
+          <para>New to Gnucash 3.0, accounts and transactions can be filtered for reporting according
+          to account full name (e.g. ":Business:" will select account structure Income:Business:Sales
+          Income:Business:Grants, Expenses:Business:Rent, Expenses:Business:Utilities) or transaction
+          description/notes/memo (select transactions tagged #gift or #travel). The filtering text
+          may be either a standard string, or a POSIX regular expression matcher.
```
jralls: POSIX regular expression. The matcher is the code that interprets the regexp.

jralls:
Could you please use more descriptive commit summaries? E.g. "Put selection items into lists." instead of "fixup progress" and "Add description of description field filter." rather than "fixup! fixup progress".

guide/C/ch_reports.xml
```diff
@@ -613,4 +613,183 @@ an empty report will be generated.</para>
...
+      <listitem>
+        <para><emphasis>New!</emphasis> Filter by Closing Transactions will disallow closing transactions (i.e. income &amp; expense accounts to equity) which should not routinely be included in transaction reports. (This is now enabled by default.)</para>
+      </listitem>
+      <listitem>
+        <para><emphasis>New!</emphasis> Options Summary will print a list of report options that were changed by the user. This is mainly useful to troubleshoot reports.</para>
+      </listitem>
+      <listitem>
+        <para>Bugs fixed - numerous bugs identified were fixed; e.g. sign display was fixed, correct grouping of periodic dates.</para>
+      </listitem>
+    </itemizedlist>
```
jralls: While we do need  a "What's new" document, it's better to make it separate--at least its own chapter if not a page on the website--rather than including it into the body of the Guide where it will get stale in a year or two.

guide/C/ch_reports.xml
```diff
@@ -613,4 +613,183 @@ an empty report will be generated.</para>
...
+  </sect2>
+  <sect2 id="rpt_TR_newbie">
+    <title>The Newbie Guide to the Transaction Report</title>
+    <para>The Transaction Report can be heavily customised to produce most reports appropriate for personal finance and business bookkeeping purposes. It aims to retrieve and filter information from the database, and present transactions and totals, useful for the user and the accountant.</para>
```
jralls: Please keep text lines to ~80 characters. It's a pain to have to side-scroll to read here; the formatter will ignore the newlines when actually creating the documents.

guide/C/ch_reports.xml
```diff
@@ -613,4 +613,183 @@ an empty report will be generated.</para>
...
+        <answer>
+          <para>Accounts: click Income, and Select Children</para>
+          <para>General: Choose absolute dates “01/07/2015” to “30/06/2016”</para>
+          <para>Sorting: Primary Key: Accounts, Secondary Key: Date</para>
+          <para>Sorting: Primary Key Subtotal: Enabled, Secondary Date Subtotal: Quarter</para>
+        </answer>
+      </qandaentry>
+      <qandaentry>
+        <question>
+          <para>I wish a list of transactions showing income transactions, grouped by day of week</para>
```
jralls: s/wish/want/

guide/C/ch_reports.xml
```diff
@@ -728,65 +728,89 @@ an empty report will be generated.</para>
...
         </answer>
       </qandaentry>
       <qandaentry>
         <question>
           <para>How much did I spend on various expenses, with monthly comparison?</para>
         </question>
         <answer>
-          <para>    As above</para>
-          <para>General: Subtotal Table enabled</para>
+          <para>The following sequence will add a subtotal summary table in a grid, which
```
jralls: "The following sequence will" repeated 3 times in succession is a bit stilted. Can you come up with a different lead in for each paragraph?

guide/C/ch_reports.xml
christopherlam: Ok let me know where to move this part to?

jralls: I think we need to discuss that on the dev list. My inclination is to put together a page for the website.

sunfish62:
Chris, Thanks for putting the effort into the documentation of this new report. I appreciate it. I am not sure how to handle suggestions for changes to your contribution. There are a number of points I raise. 

My question is, how might I go about offering my editorial skill on this? I'd be happy to offer such input.

Here are several issues:
1) Is the new report replacing the original Transaction Report, or is it in addition to it? If it is a new report, then it needs a *different* name from the existing Transaction Report. If it replaces the old report, then your text should *replace* the text currently in place for the Transaction Report (at line 581-586). Regardless, I don't think this report should be handled in a completely different section of the chapter--it should be another description of a report that is included with GnuCash. This will affect the markup.
2) I prefer NOT making version claims in the documentation. Adding such notes in the documentation means that future editors (such as they might be) must remember to go through and remove these references once they become stale. This is not something that the GnuCash community is particularly adept at, so it is better to avoid it altogether. I believe that John Ralls has suggested putting something into the website or Wiki to explain this new report, and I think that is a far better approach.
3) Stylistically, I would prefer that "sales-pitch" language be removed. Technical writing should emphasize what the software does; marketing writing (e.g., on the website or wiki) is where such promotional language should be used. Thus, every "New!" should be (IMO) be removed. Similarly, users years from now aren't going to care about bug fixes; they are only going to want to understand what the report is capable of doing.
4) A personal stylistic concern is the use of "Newbie". To my ear, this is rather derogatory, and I have attempted to iron such language out of the official documentation. Again, a focus on clear writing that outlines the functions and features should be able to avoid this terminology.
5) I agree that the Q&A format is not right for this. It would be better to configure it as explanatory description rather than Q&A.

christopherlam:
Hi David

Sure, feedback welcome :) I was writing on Google Docs but I figure writing in a Pull Request will attract the right amount of attention.

> Is the new report replacing the original Transaction Report, or is it in addition to it? If it is a new report, then it needs a different name from the existing Transaction Report. If it replaces the old report, then your text should replace the text currently in place for the Transaction Report (at line 581-586). Regardless, I don't think this report should be handled in a completely different section of the chapter--it should be another description of a report that is included with GnuCash. This will affect the markup.

The report is definitely replacing the original Transaction Report which was a Frankenstein monster of hooks, patches and dead code. I'd hope the new TR is much easier to understand, even to a non-programmer. Let me know where texts should belong.

> I prefer NOT making version claims in the documentation. Adding such notes in the documentation means that future editors (such as they might be) must remember to go through and remove these references once they become stale. This is not something that the GnuCash community is particularly adept at, so it is better to avoid it altogether. I believe that John Ralls has suggested putting something into the website or Wiki to explain this new report, and I think that is a far better approach.

NP awaiting further outcome. There are definitely functional changes which will affect saved reports or anyone who programmatically processes the reports, therefore I felt they should be documented and highlighted.

> A personal stylistic concern is the use of "Newbie". To my ear, this is rather derogatory, and I have attempted to iron such language out of the official documentation. Again, a focus on clear writing that outlines the functions and features should be able to avoid this terminology.

"A Quickstart  Guide to the Transaction Report" ?

> I agree that the Q&A format is not right for this. It would be better to configure it as explanatory description rather than Q&A.

The purpose of Q&A is to solve common scenarios -- after all the TR is still a monster report program with lots of ways to generate nonsense reports e.g. primary sortkey = amount, secondary sortkey = reconcile date, filter = void only, will be completely silly. Hence, gently introducing common uses. A drier technical description can be written in the Help section.

sunfish62:
OK. New idea. Since you describe this report as "a monster", perhaps it would be more appropriate to put its description at the top of the Reports chapter, in its own special section, as a special case report. Lines 581-586 would be deleted. Your stuff would be put in at line 79. Lines 75-78 would be changed to read as below, and your piece would go in after:

<para>The standard reports that are included in <application>&app;</application> are presented in this chapter. For each report, a short description is given, which explains what the report is intended to show and its primary purpose. We begin with a description of the Transaction Report, which is an all-purpose report with many features and settings that allow sophisticated examination of the data in a file. </para>

<sect2 id="rpt_txnrept">
  <title>Transaction Report</title>
   <para>The Transaction Report is a fundamental report that allows users to retrieve
    a wide variety of useful information from the financial records. The report includes many features that allow powerful and flexible reporting from the single basic report.</para>
    <itemizedlist>
      <listitem>
        <para>Debit and Credit subtotals can be displayed, and within a grouping section, if
        debit totals > credit totals, the group subtotal will be displayed in
        the debit column.</para>
      </listitem>
... etc.

As for the "New and Improved" language, there is discussion on the devel list to create a webpage for this, and I think that can alert users sufficiently. I am not sure what you mean by "anyone who programmatically processes the reports," but my general sense is that such users (if they exist) are going to be savvy enough to examine the implications of your changes. Effects on saved reports might be handled in the release announcement.

As for titling, in the context of the Guide, there isn't any need for even "A Quickstart Guide to the Transaction Report." The guide is the Guide, after all, and your section is rather detailed, giving the lie to the idea that it is in any way a "Quickstart". ;)

Finally, here is how I might handle one of your Q&A entries, added perhaps in a separate "Using the Transaction Report" section:

<title>Using the Transaction Report to Show Previous Year Expenses</title>
<para>The Transaction Report can show how much was spent on expense accounts last year. To create this report:</para>
<orderedlist>
<listitem><para>Open the report options, select the Expense account and its children.</para></listitem> 
<listitem><para>Set the date options to “Start of Previous Year” and “End of Previous Year.”</para></listitem> 
<listitem><para>Open the Sorting tab, set Primary Key to “Account” and enable Primary Key Subtotal.</para></listitem> 
<listitem><para>Secondary Key to “Date” and set Secondary Key Subtotal to None.</para></listitem>

christopherlam:
Due to string freeze I'm not sure if 'subtotal-grid' or 'weekdays' will be ready for 3.0 so these will need to be reviewed prior to release. I haven't yet written the drier technical report in Help section.

fellen:
At least you can use the strings defined in (maint)/src/gnome-utils/gtkbuilder/gnc-frequency.glade.h.
Obey that our slavian friends (ru, ...) need some context for the right translation.
You should get the allowed full list by 'grep "msgid" gnucash.pot' after make pot in remote/unstable's .build/po.

sunfish62:
Chris, this is much better. Thank you! The writing is much cleaner.

I'm not fully convinced on the separate Using section, but I think it's better to get the information out there, and then see what needs changing later. 

At this time, I think my only additional request is to add an internal link from the introductory section down to the Using section at the bottom. Something as simple as "For ideas on how to use the Transaction Report, see {link-to-new-section}."

christopherlam:
Interesting... we can try access gncFrequency strings from scheme.

guide/C/ch_reports.xml
```diff
@@ -121,6 +121,9 @@ Configuration.</para>
         (This is now enabled by default.)</para>
       </listitem>
     </itemizedlist>
+    <para>For further ideas on how to use the Transaction to produce
```
sunfish62: should be "Transaction Report"

guide/C/ch_reports.xml
```diff
@@ -613,4 +655,216 @@ an empty report will be generated.</para>
 </sect3>
 </sect2>
 </sect1>
+
+<sect1 id="rpt_transaction">
```
sunfish62: Would this be better placed as a final sect2 under sect1 rpt_standardrpts?

guide/C/ch_reports.xml
```diff
@@ -1,10 +1,11 @@
 ﻿<!--
       (Do not remove this comment block.)
   Version: 2.6.15
```
fellen: Do not forget to update the version number at least to the minimal required one
* in both files
* when everything else is OK

gjanssens:
What's the status of this ? Can it be committed or do the last to WIP commits need review ?

christopherlam:
Finalise with features that have been merged into Gnucash 3.0. Last review requested!

guide/C/ch_reports.xml
```diff
@@ -755,29 +755,38 @@ an empty report will be generated.</para>
...
-        <listitem><para>From the <emphasis>Sorting</emphasis> tab,
-        set <emphasis>Primary Key</emphasis> to Accounts, and enable
-        <emphasis>Primary Key Subtotal</emphasis>.
-        </para></listitem>
-        <listitem><para>Set the <emphasis>Secondary Key</emphasis> to
-        Date, and set the <emphasis>Secondary Key Subtotal</emphasis>
-        to None. </para></listitem>
+        <listitem>
+          <para>Open the report options, from the
+          <emphasis>Accounts</emphasis> tab, click the Expense
```
fellen: `<guilabel>Accounts</guilabel>`

pls. check other occurrences of `<emphasis>`.
A incomplete list of gui elements:
accel, guibutton, guiicon, guimenu, guimenuitem, guisubmenu, keycap, keycode, keycombo, keysym, menuchoice, mousebutton, shortcut.

See e.g. https://tdg.docbook.org/tdg/5.0/docbook.html

guide/C/ch_reports.xml
```diff
@@ -107,18 +110,15 @@ Configuration.</para>
       </listitem>
       <listitem>
         <para>Indenting of columns will indent the report columns for
-        easier understanding of grouping and sorting. (This
-        is now enabled by default)</para>
+        easier understanding of grouping and sorting. (This is now
```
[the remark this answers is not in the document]
gjanssens: This is a new feature for 3.0 IIUC. I wouldn't add version information because when reading this guide it should be clear it's the 3.0 guide.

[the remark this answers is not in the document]
fellen: That is OK for the first time reader, but I think a person which already has read the 2.x guide should have some clue something changed. Probably there were better approachs but they would need too much time.

PR #107, opened 2018-05-17
Progress

christopherlam:
1. Add Closing Txn Filter to help
2. Add sample tax tables to guide


sunfish62:
On line 558 of ch_help_reports: this should read "excluding" instead of "exclusing."
Note: If the preferences in the program say "exclusing," then this would need to be corrected in code to use "excluding", since that is the real word.

gjanssens:
Pulled, thanks!

PR #110, opened 2018-08-24
Adding reconciliation to glossary

sunfish62:


gjanssens:
Thanks David to run the experiment of working directly on github.

That allows me to write my review there as well :)

I have two remarks:

1. We generally ask "commits" to reference the bug they fix if there is one. I see you have chosen to reference the bug in your branch name instead. The best way to do this is to use the bug and bug title as commit title (the first field in the "Commit changes" frame on the edit page).
Any further clarifications or comments can be added in the second field.
1. Your PR is crossing branches. That is, you created your commit starting from the maint branch (good, as this change is useful for gnucash 3.x and up) and then created a PR against master. That should be avoided.

So even though the github interface is cleaner a minimal understanding of git branches is still needed when unsing the integrated editor. This is in no way meant to comment on your effort. Rather I'm using your experiment to draw conclusions and discover pitfalls.

Do you want to continue the experiment and see if you can correct this ?
I don't think you can change the commit message unless you redo the commit. You don't have to, I'll do so when pulling your PR.
However you can test how hard you feel it is to fix the PR to be against the proper branch. If you want to, the way to do so is to close this PR, go back to your "Bug-791169---Add-Reconciliation-definition" branch and create a new PR, this time against the maint branch.

gjanssens:
Oh and perhaps another pitfall: I currently haven't found a way to synchronize your personal fork with the main gnucash-docs repository directly on github. This doesn't happen automatically as far as I know so if you have your fork around for some time it may be getting out of sync and you may end up creating PR's from outdated information.

Perhaps there is a way to synchronize, but I haven't found it yet.

sunfish62:
At your behest, I am closing this pull request as being against Master instead of Maint. Github seems to like changing reference points, which requires a user to pay close attention to them. In this example, I can't imagine why I'd make changes on Maint, and then want to issue the PR against Master.

fellen:
@gjanssens: I believe **owners** can change the default branch. GitHub seems to set the default to master, but if I open https://github.com/Gnucash/gnucash, I get branch maint. IMHO we should also on gnucash-docs and probably other repos like for packaging, but not for ht-docs, set the default branch to maint because that is the most often used. Because I am a simple **member** I seem not be able to do it.

gjanssens:
Good point. I have changed this on gnucash-docs. gnucash-on-osx and gnucash-on-windows only have  a master branch so there's nothing the do there. Finally there's gnucash-on-fedora-copr, my attempt to get nightly builds for Fedora. I am about to retire that one. I think flatpak/snap/whatever container is a more viable option for linux nightlies (on my list of itches...).

PR #111, opened 2018-08-24
Bug 791169 - Add Reconciliation definition to glossary

sunfish62:
Welp. Maybe this one will go the right way

### DIFF
--- a/guide/C/gnc-glossary.xml
+++ b/guide/C/gnc-glossary.xml
@@ -318,6 +318,18 @@
   </glossdef>
 </glossentry>
 
+<glossentry id="gnc-gl_reconciliation">
+  <glossterm>Reconciliation</glossterm>
+  <glossdef>
+    <para>Reconciliation is a verification process in which the user compares their records
+    with those of the financial institution to ensure that the records in each system agree 
+    at some particular point in time. During reconciliation, the user checks individual 
+    transactions in their file against the institution's statement; upon completion, the 
+    reported balances of the institution and the reconciled account will match at that 
+    point in the register.</para>
+  </glossdef>
+</glossentry>
+
 <glossentry id="gnc-gl_return">
   <glossterm>Return</glossterm>
   <glossdef>


gjanssens:
It did :)
Thanks!

If you want you can now delete the two branches related to this bug from your own fork. We still need to find a way to update your local fork so it will pull in these changes in its maint branch as well.

derekatkins:
https://help.github.com/articles/syncing-a-fork/
Seems to imply you need to do it via a local sync spot.

BUT, googling seems to imply there is a way to do it online:
https://github.com/KirstieJane/STEMMRoleModels/wiki/Syncing-your-fork-to-the-original-repository-via-the-browser

Alas, it is still a multi-step process.  In short, you create a pull
request to yourself from the upstream repo.

-derek


On Fri, August 24, 2018 1:30 pm, Geert Janssens wrote:
> It did :)
> Thanks!
>
> If you want you can now delete the two branches related to this bug from
> your own fork. We still need to find a way to update your local fork so it
> will pull in these changes in its maint branch as well.
>
> --
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/Gnucash/gnucash-docs/pull/111#issuecomment-415827786


-- 
       Derek Atkins                 617-623-3745
       derek@ihtfp.com             www.ihtfp.com
       Computer and Internet Security Consultant

gjanssens:
Yes, I had figured out the PR to your own repo method already.

Unfortunately this means that in order to make regular contributions to the documentation via github one still needs to understand the git concepts fairly well. On top of that PR's generated directly on github did not even get the basic syntax check for which we currently ask the contributors to set up a build chain locally.

So yes, the user doesn't have to bother with it, but the PR is not syntax checked either.

So far these are my conclusions of this experiment:

1. It is possible to make contributions directly via github. It can save the contributor the effort to learn how to work with git on a local machine.
2. On the other hand it doesn't hide much of git either:
- One still has to fork the upstream repo
- One still has to understand the concept of branches. Know which branch to start from, know how to create a new branch based on it, know which branch to make a PR against...
- One still has to understand how to make proper commit messages (title should contain the bug number and summary, more details go into the additional field) and github is overzealous in guiding the user here, which leads to uncertainty.
- One still has to learn how to keep the forked repo in sync with the upstream repo. And github makes this actually more difficult.
3. Drawbacks:
- Already mentioned: it's at least equally difficult if not more so to keep your personal repo in synch with the upstream one.
- Working directly in github forgoes the syntax checks.

We can alleviate the burden of syntax checks by configuring a travis run for documentation, similarly to what we do for code. That way we can be sure PR's are checked before merging and at the same time remove the burden for documentation contributors to learn development tools. The documentation update process would then include the requirement to check the CI result on the PR instead of running make check. Of course interpreting that result remains.

The other drawback - keeping in sync without requiring a local repository - is more of a challenge. For one-off contributions this probably is not an issue. Such a contributor would fork the repo as it is, make the PR and be done with it. For longer-term contributors this becomes more difficult. One clumsy workaround could be to create a new fork for each topic instead of a new branch for each topic. Or at least start a new fork regularly to get the most recent upstream changes. Older forks can only be deleted when all PR's from it are merged or closed. That way the contributor doesn't have to be concerned with pulling in upstream an potential merge conflicts.

Another solution would be to give each interested contributor access to the upstream repo and have them work directly on the documentation sources in there online. Breakage is not a serious problem as we can always revert bad commits. The main issue is though that the github repo is only a mirror of the real upstream: code.gnucash.org. So we can't do this on github. If we had our own gitlab instance on the other hand...
But even in that case, I think the contributor experience improvements would only be marginal, at the potential expense of an increase maintenance burden.

So to conclude: a direct contribution via github is viable, but has several drawbacks of its own. I can't estimate which method would be preferred by non-dev contributors. 

So back to @sunfish62 : 
* How would you ensure that your own fork is up to date with our primary repo without loosing your current unmerged word ?
And with that in mind, if we can eliminate the make related dependencies from the documentation process, would working locally be more or less appealing to you than working directly on github ?

PR #113, opened 2018-08-31
Bug 777893 - Adding Edits

sunfish62:
These are editorial changes based on discussion. Not sure if this is how this is supposed to be handled...

### DIFF
--- a/guide/C/ch_basics.xml
+++ b/guide/C/ch_basics.xml
@@ -643,6 +643,7 @@
 		<guimenuitem>Save As...</guimenuitem></menuchoice>. This will create a copy of 
 		the data file in the selected format.</para>
 		</note>
+	</sect2>
 
 	<sect2 id="basics-files-storage-comparison">
 		<title>Storage Comparison and Recommendations</title>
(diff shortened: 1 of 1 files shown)


gjanssens:
You're doing fine except you don't have to create a new pull request. Just pushing your changes into your private github repo to the same branch will update the existing pull request.

I'll close this one as most of the discussion has happened so far on PR#109. Let's keep it there.

(For completeness I have noted PR#109 is against master while it should be against maint like in this PR. However in practice I can also merge PR#109 into maint without requiring you to change anything, so you can ignore that part.)

fellen:
This is the continuation of PR #109. 
@gjanssens It might be the easiest way to get rid of commit 93dbe52 (glossary) and get the right branch.
@sunfish62: The downside: Overlapping I commented on PR #109. What is your opinion about my last review there?

fellen:
@gjanssens: If you put a space between "PR" and "#" you will get a reference in the other PR.

gjanssens:
Well given this PR is made of the exact same branch (though against the proper upstream branch), it has the same unwanted merge commit. So I will need to do some git magic regardless of the branch we choose. For me it's a small correction any way.

And as most of the discussion was done on PR #109 I decided to stick with PR #109 and close this one.

PR #114, opened 2018-09-15
Bug 796856 Help Ch6 section added for importing from files

DaveC49:
Incorporates changes associated with Bug 796778 re multiple selections in import matcher I could n't find anywhere in the Help or Guide to document the above so I have added a section to the Help manual inCh 6  Transactions on Importing files before the stub for Import Online Actions. I copied the Import QIF section from the wiki. Someone might like to check i haven't made any glaring errors.

### DIFF
--- a/help/C/Help_ch_Transactions.xml
+++ b/help/C/Help_ch_Transactions.xml
@@ -859,6 +859,267 @@ Translators:
     to print the check.</para>
 
   </sect1>
+  <sect1 id="trans-import">
+    <title>Importing Transactions from Files</title>
+    
+    <para>Imported transactions will generally be to a specific account in your account tree. In the following 
+            this will be referred to as the <emphasis role="bold">import</emphasis> or 
+            <emphasis role="bold">base</emphasis> account. It may or may not be
+            specified in the data being imported depending on the import format. It is usually the first split
+            of a transaction being imported.
+    </para>
+    <para>All transactions will also need to have a <emphasis role="bold">destination</emphasis> account for 
+            at least a matching splits. This may or may not be supplied in the imported data. If it is not,
+            an account can be assigned on the basis of the previous import history by matching to infomation in
+            the imported data. The user may always over-ride this assignment.
+    </para>
+    <para>Multi-split data previously exported from GnuCash may have both the import and destination accounts
+            for transaction splits specified in the data file.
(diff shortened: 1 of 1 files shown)


sunfish62:
I'll add here the comment I made on gnucash-devel, that this looks generally good, except that the list of import formats needs to mention all formats (even if detail isn't provided at this time), and that this would be better placed in the new Importing chapter of the Guide.

fellen:
Before it gets bit rotten - we missed already 2 releases -, I apply it now. It is an improvement. :-)

@DaveC49 
1. I wonder, if we should join Import and Export to Data Exchange as several formats can be used in both directions, the german Dataus format and FinTS methods also to send orders to the bank.
1. Probably we should clearer distinguish "creating a new file" by importing (distinct accounts) and "updating an existing account". 
1. `<listitem><para><emphasis role="bold">Encoding</emphasis> ...`: We should generally avoid hard formating. That will be done by xsl. In this case we have `<listitem><para><guilabel>Encoding:</guilabel> ...` until Bug 633673: "In DocBook sources, ItemizedLists should be VariableLists" is fixed. See "User interfaces" in [Creating DocBook Documents](https://tdg.docbook.org/tdg/4.5/ch02.html).
1. Can you add links to the new sections in the table AccTree-FileMenu in ch_GUIMenus?

@sunfish62 

> I'll add here the comment I made on gnucash-devel, 

Link?
> that this looks generally good, except that the list of import formats needs to mention all formats (even if detail isn't provided at this time), 

Right,
-  google SWIFT MessageTypes. Links to the respective wikipedia articles might be sufficient.
- After 40 years of usage DATAUS is no longer official supported by german banks and will be dropped with the next AqBanking release.
> and that this would be better placed in the new Importing chapter of the Guide.

I am not sure. At least the descriptions of the file formats are a technical details, which belong in the manual,

BTW, both of you should add their `<author><firstname>...</firstname> <surname>...</surname></author>,` section to the main documents of the touched parts.

fellen:
In commit 06ca2f3 I have updated GUImenus from Aqb to give you some clue about MT9xx, DTAUS, ...

fellen:
I almost forgot to mention: In commit f92b816 I replaced most occurrences of <emphasis> in help:transactions.

PR #120, opened 2019-03-26
Itstool experimental branch

gjanssens:
This branch is meant to experiment with an alternative workflow for documentation authoring and translations. In this alternative workflow the independent documents per language are replaced with a single master document and translation will happen using gettext and the its extensions.

The single master document will serve for all languages as follows:

* Sections with contents that is valid for all languages will be written in English.
* Using the proper tooling the translatable messages will be extracted into a message catalog (pot file) which can be translated for each language we support.
* Sections that are relevant only for a single language will also  be added in the master document, but marked with special ITS tags. As a result messages from these sections will not appear in the catalog files.
* ITS also supports sections that are valid for some but not all languages or sections that are not relevant for some languages.

This branch is working with a reduced sample book (many chapters removed) for simplicity.

It has converted this subset into a master document and has created a po file for the German language.

Note creating this initial PO file is fairly complicated (the *pot* file on the other hand is easy). I have written a script that will extract msgids and corresponding msgstrs from the original and translated files. The only caveat is the order in which the xml tags appear in both documents has to match exactly. If they don't the script will fail. So to use the script any tag mismatches should first be manually corrected (using a smart diff tool to find the differences does help, but it's still a huge one-time manual effort).

Lastly what gettext with ITS won't solve for us is that incomplete translations will be replaced with English text as opposed to not included as it was in the original workflow. I'll admit that personally I would prefer complete documentation that may be partially untranslated than incomplete documentation with what's there fully translated. Including untranslated contents may also better expose the work to be done still to potentially new translators. We can investigate how tooling/scripting can improve the experience though.

Note also that the last commits are there to simulate large documentation changes and evaluate how this impacts translatability of the master document and the language specific po files in particular.


pmraps:
Bravo!
As usual, an excellent idea. Looking forward to see it working.

fellen:
I hope, after the 3.5 release to have some time for the points already mentioned in https://wiki.gnucash.org/wiki/Po_Based_Documentation_Translations

Others can also share their knowledge there. ;-)

PR #122, opened 2019-05-30
Draft Debit and Credits section

DaveC49:
Debit and Credit section showing which accounmt types have debit balances and which haave credit balances and the effect of debits and credits on each account type.
Debits and Credits section

Minor fix to missing tag
Double Entry section

Add a reference to the debit and credit section above

### DIFF
--- a/guide/C/ch_basics.xml
+++ b/guide/C/ch_basics.xml
@@ -142,7 +142,49 @@
         </screenshot>
       </figure>
     </sect2>
-
+    
+    <sect2 id="basics-debits-credits">
+      <title>Debits and Credits</title>
+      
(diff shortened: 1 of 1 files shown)


DaveC49:
Section 2.1.3  Debits and Credits explaining the conceptual basis of debit balance accounts and credit balance accounts and how to determine which is which from the accounting equation added

fellen:
Sorry for the delay.

This PR relates to Bug 797265 - Basics - Explanation of Debits and Credits

it is a very good section, thank you!

I found only one typo, added an ID and used <equation> instead of <emphasis> as you can see in my follow ups.

DaveC49:
Thanks Frank.

PR #123, opened 2019-07-10
Import Bills & Invoices: blank ID is replaced by ID from previous row

ekinonnakapito:
In the Import Bills & Invoices function, previously a blank invoice ID was considered an error. After the latest changes in Gnucash pull request #457, a blank invoice ID is substituted with the ID of the previous row. So a blank ID is only an error in the initial row of an import file.


fellen:
Sorry, this was forgotten.
Thanks!

PR #124, opened 2019-07-14
Optimize all PNGs with optipng

SkewedZeppelin:
Reduces size by ~28MB

should be beneficial to users, and distro mirrors alike.


fellen:
I think, it is a very interesting idea. A few notes without testing or looking in details like "What is the meaning of IDAT?":

- Did you use specific parameters? How long was it running? Can attach the log file?

- Should we document the use in a section like https://wiki.gnucash.org/wiki/Documentation_Update_Instructions#Adjusting_an_Image.27s_Dots_Per_Inch? As alternative one could make it part of the build process, but I think it would be too time consuming.

- I am not sure we should apply it on borrowed parts like stlesheet/ and xsl/

SkewedZeppelin:
> Did you use specific parameters?

Defaults.

> How long was it running?

About 15 minutes from RAM (/tmp)

> Can attach the log file?

That was very long (past terminal scrollback of 10k lines) and never saved

> Should we document the use in a section

Yes, a section to run optipng (or any other optimizer) on images on check-in would be beneficial

> I am not sure we should apply it on borrowed parts like stlesheet/ and xsl/

I can drop those and force push if you want

> The notification mail was incomplete

Mail? From GitHub or a bridge to a list?

fellen:
Thanks, I have applied it and adjust now https://wiki.gnucash.org/wiki/Documentation_Update_Instructions, also with the infos from your comment-

PR #125, opened 2019-07-15
Modify Matcher and Multiple Selection Docs

DaveC49:
Changes to improve the description of the import matcher operation and improve the section on the multiple selction of transactions to reflect changes incorporated in v3.6
add a table table of matcher status conditions.

### DIFF
--- a/guide/ru/fop.xconf
+++ b/guide/ru/fop.xconf
@@ -1,0 +1,58 @@
+<?xml version="1.0"?>
+<fop version="1.0">
+  <font-base>/home/david/GnucashDevelopment/Docs/gnucash-docs/./fonts</font-base>
+     <!-- Source resolution in dpi (dots/pixels per inch) for determining the size of pixels in SVG and bitmap images, default: 72dpi -->
+  <source-resolution>72</source-resolution>
+  <!-- Target resolution in dpi (dots/pixels per inch) for specifying the target resolution for generated bitmaps, default: 72dpi -->
+  <target-resolution>72</target-resolution>
+  
+  <!-- page-height and page-width, A4 for Russian -->
+  <default-page-settings height="297mm" width="210mm"/>
+  
+  <!-- Information for specific renderers -->
+  <!-- Uses renderer mime type for renderers -->
+  <renderers>
+    <renderer mime="application/pdf">
+      <filterList>
+        <!-- provides compression using zlib flate (default is on) -->
+        <value>flate</value>
+      </filterList>
+
+      <fonts>
+        <font embed-url="opentype/freefont/FreeSerif.otf">
+          <font-triplet name="serif" style="normal" weight="normal" />
+        </font>
+        <font embed-url="opentype/freefont/FreeSerifBold.otf">
+          <font-triplet name="serif" style="normal" weight="bold" />
+        </font>
+        <font embed-url="opentype/freefont/FreeSerifItalic.otf">
+          <font-triplet name="serif" style="italic" weight="normal" />
+        </font>
+        <font embed-url="opentype/freefont/FreeSans.otf">
+          <font-triplet name="sans-serif" style="normal" weight="normal" />
+        </font>
+        <font embed-url="opentype/freefont/FreeSansBold.otf">
+          <font-triplet name="sans-serif" style="normal" weight="bold" />
+        </font>
+        <font embed-url="opentype/freefont/FreeSansOblique.otf">
+          <font-triplet name="sans-serif" style="italic" weight="normal" />
+        </font>
+        <font embed-url="opentype/freefont/FreeMono.otf">
+          <font-triplet name="monospace" style="normal" weight="normal" />
+        </font>
+        <font embed-url="opentype/freefont/FreeMonoBold.otf">
+          <font-triplet name="monospace" style="normal" weight="bold" />
+        </font>
+        <font embed-url="opentype/freefont/FreeMonoOblique.otf">
+          <font-triplet name="monospace" style="italic" weight="normal" />
+        </font>
+      </fonts>
+
+      <!-- This option lets you specify additional options on an XML handler -->
+      <xml-handler namespace="http://www.w3.org/2000/svg">
+        <stroke-text>true</stroke-text>
+      </xml-handler>
+
+    </renderer>
+</renderers>
+</fop>
--- a/help/C/Help_ch_Transactions.xml
+++ b/help/C/Help_ch_Transactions.xml
@@ -1067,28 +1067,112 @@ Author:
     </sect2>
     <sect2 id="trans-import-matcher">
       <title>Import Matcher</title>
-      <para>The Import Matcher uses a Bayesian approach to assign a destination account, if it is not specified in 
+      <para>The Import Matcher uses a Bayesian approach to assign a transfer account, when it is not specified in 
           the imported data, to each imported transaction based on the previous import history of the import account.
-          It also attempts to match the transactions being imported to any existing transactions
-          based on the date and the description fields<!-- FIXME add any other fields used in matching -->.
+          It also attempts to match the transactions being imported to any existing transactions already recorded
+          based on the date, amount and the description fields by tokenizing that information and comparing it to stored tokens for 
+          assigned transfer accounts assigned in previous imports.<!-- FIXME add any other fields used in matching -->.
       </para>
-      <para>Transaction rows which match existing transactions already in the import account are flagged not to be imported. They 
-        will have a light green background and the A and U+R checkboxes will be unchecked and the R checkbox will be 
-        checked. To override and import the transaction, check the A checkbox. The U and R boxes will be unchecked 
-        automatically. The reliability of the match is indicated by a bar display in the Info column. If a destination
-        account for the second split is assigned by the matcher if will be appended to the info column.
+      <para>The status of the imported transaction rows is indicated by a combination of the background colour for the row and three
+       checkboxes in the columns labelled A (Add), U+R (Update and Reconcile) and R (Reconciled) as described in <xref linkend="match_status"/>.
       </para>
-      <para>Transaction rows which do not match existing transactions in the import account, for which an assignment
-        of a destination account cannot be made on the basis of the previous import history to the account, will be 
-        displayed with an orange-yellow background and the A box will be checked and U+R and R unchecked. A destination 
-        account must be specified for these transactions.
-      </para>
+      <note><para>Note that reconciled in this context is not the same as reconciled in the process of reconciling an account.</para></note>
+      <table id="match_status" frame='all'><title>Status of transactions in the Import Matcher</title>
+        <tgroup cols='3' align='left' colsep='1' rowsep='1'>
+        <colspec colname='c1'/>
+        <colspec colname='c2'/>
+        <colspec colname='c3'/>
+        <thead>
+          <row><entrytbl cols='2' colsep='1'><tbody><row><entry>Checkbox</entry><entry>State</entry></row></tbody></entrytbl><entry>Row Background Color</entry><entry>Meaning</entry></row>
+        </thead>
+        <tbody>
+          <row>
+            <entrytbl cols='2' align='left' colsep='1' rowsep='1'>
+                <tbody>                  
+                  <row><entry><emphasis role="strong">A</emphasis></entry><entry><emphasis>X</emphasis></entry></row>
+                  <row><entry><emphasis role="strong">U+R   </emphasis></entry><entry>-</entry></row>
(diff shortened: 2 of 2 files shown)


guide/ru/fop.xconf
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0"?>
+<fop version="1.0">
+  <font-base>/home/david/GnucashDevelopment/Docs/gnucash-docs/./fonts</font-base>
```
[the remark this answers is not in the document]
DaveC49: Hi Frank will do. I used an the XMLmind editor to edit the sources so that must have introduced the font. Strange that it into the document. Useless editorif it does that.

David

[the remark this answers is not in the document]
DaveC49: Frank,

That has now beenremoved and the change pushed through to my repository. I don't think it was the XMLMind editor though as I did no editing of anything other thaHelp_ch_Transactions.xml.  I did initally build all the documentation and the build obviously picked it up from my computer. Presumable a make clean would have got rid of it as well.  I guess that is the problem with the in source build with autotools.
Cheers
David

help/C/Help_ch_Transactions.xml
```diff
@@ -1067,28 +1067,112 @@ Author:
...
-        account for the second split is assigned by the matcher if will be appended to the info column.
-      </para>
-      <para>Transaction rows which do not match existing transactions in the import account, for which an assignment
-        of a destination account cannot be made on the basis of the previous import history to the account, will be 
-        displayed with an orange-yellow background and the A box will be checked and U+R and R unchecked. A destination 
-        account must be specified for these transactions.
+      <para>The status of the imported transaction rows is indicated by a combination of the background colour for the row and three
+       checkboxes in the columns labelled A (Add), U+R (Update and Reconcile) and R (Reconciled) as described in <xref linkend="match_status"/>.
       </para>
+      <note><para>Note that reconciled in this context is not the same as reconciled in the process of reconciling an account.</para></note>
```
fellen: This needs a better explanation. Isn't it becoming "cleared" in the reconcile dialog?
@jralls should it be changed to from R/Reconciled to C/Cleared in the dialog + updates in other parts of the documentation like screenshots?

help/C/Help_ch_Transactions.xml
```diff
@@ -1067,28 +1067,112 @@ Author:
...
+                <row><entry>Matches an existing transaction - not to be imported.</entry></row>
+                <row><entry>Matches but not reconciled - not to be imported.</entry></row><!--FIXME might not have this quite correct -->
+                </tbody>
+            </entrytbl>
+          </row>
+          <row><entry></entry><entry><?dbhtml bgcolor="#6699CC"?><?dbfo bgcolor="#6699CC"?>Cornflower Blue</entry><entry>Row is selected</entry>
+          </row>
+        </tbody>
+        </tgroup>
+       </table>
```
fellen: Table match_status in yelp 3.32.2
I believe, you intended something different?

help/C/Help_ch_Transactions.xml
```diff
@@ -1067,28 +1067,112 @@ Author:
...
+       <para><note>The colors illustrated here are the defaults for GnuCash on a Linux system. If you have altered the CSS
+        or are using another OS the colors may be different.</note>
```
fellen: I am not  sure, which colors are hardcoded and which are inherited from your desktops default theme. I am now too lazy to look in the source.

guide/ru/fop.xconf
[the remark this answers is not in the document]
fellen: Yes, it is a typical artefact, if you build in-tree.

But now I am confused by your merge (3. commit). I will cherry-pick your clean 2. commit.
I will add a few more comments, but because of the merge, I will close this PR and ask you to improve the text/table in separate PRs.

help/C/Help_ch_Transactions.xml
jralls: @fellen, Do you mean should we change the heading in the dialog from U+R to U+C? If that's what it really does, I agree that we should.

guide/ru/fop.xconf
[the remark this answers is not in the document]
DaveC49: OK No problem Frank

help/C/Help_ch_Transactions.xml
DaveC49: With the OFX and CSV imports it certainly marks the transactions as "c" in the R column in the register for new records imported which aren't matched. I am not sure what  occurs if an update to an existing transaction which is already reconciled and marked as such by the reconciliation procedure for an account. When the transaction is flagged as R in the matcher it really only implies that the transaction has been matched so maybe an "M" and "U+M" may be more appropriate even that the "C" and "U+C". I would guess that where an import updates an existing transaction which was already reconciled, it should undo the reconciliation status "y" in the register and set it back to "c", particularly if it was the amount which was updated but even if a description was changed since reconciliation really implies that you are satisfied that the transaction in the external statement and the transaction recorded in Gnucash are one and the same, not just that the balances are right at the end date of a reconciliation.

help/C/Help_ch_Transactions.xml
DaveC49: @fellen Certainly not what I intended. Yelp seems to handle the nesting of tables differently from HTML. I'll check through the various other output formats and see if I can get a form of the table which comes up correctly in all of them. I should be able to specify colum widths to get the header titles to match up with the columns in the nested tables.

help/C/Help_ch_Transactions.xml
DaveC49: AFAIK the register colors in GnuCash aren't inherited from the desktop theme as I have changed that without affecting the GnuCash colors at least on Linux Mint and hopefully that should be the case on most Linux distributions.  The default colors are the same on Windows10 ( I ran GnuCash on my wifes Windoze box to check that out). It is more likely to be a problem if someone has customized the GTK CSS to assign different colors. That really can't be controlled, which was why I felt it was worth noting the colors may be different just to alert anyone who does modify the colors.  I selected the colors I used in the table  here by matching the displayed default colors in GnuCash to those from a table of standard internet colors displayed on my monitor in a side by side match. I have tried to track down where the colors are allocated in the code without too much success as yet. They seem to be supplied by gtk_style_context_get() which presumably has defaultsset and also checks any  gtk CSS file. I will keep searching.

Bob-IT: @DaveC49 Have a look in gnucash-fallback-310.css

help/C/Help_ch_Transactions.xml
```diff
@@ -1067,28 +1067,112 @@ Author:
...
+           <entrytbl cols='1'>
+                <tbody>
+                <row><entry><?dbhtml bgcolor="#FFCC00"?><?dbfo bgcolor="#FFCC00"?>Yellow-Gold</entry></row> <!-- FIXME need to check the color values against those used in GnuCh -->
+                <row><entry><?dbhtml bgcolor="#CCFFCC"?><?dbfo bgcolor="#CCFFCC"?>Light Green</entry></row>
+                </tbody>
+            </entrytbl>
+           <entrytbl cols='1'>
+                <tbody>
+                <row><entry>Flagged for input-transfer account must be selected.</entry></row>
+                <row><entry>Flagged for input-transfer account matched.</entry></row>
```
fellen: How about typographical correct `Flagged for input&mdash;transfer account...` here and later
or more modern `Flagged for input &ndash; transfer account...`
or `Flagged for input - transfer account...` like below?

DaveC49: Frank,

I will create the further mods in a new branch as you suggested earlier and submit a new PR when they are done.  Bob has provided a reference to the default gtk CSS which will help with getting the right default colors. It may take me a while to get the table structured so it comes up properly in all the various formats.

help/C/Help_ch_Transactions.xml
DaveC49: @jralls  If you think the above changes make sense and won't be too confusing for long term users, I am happy to make the changes them in the code and then the documentation to match. Would it be better perhaps to raise this as a bug to facilitate any discussion/
David

help/C/Help_ch_Transactions.xml
DaveC49: @fellen  The problem with the yelp viewer is an inability to render docbooks properly. The docbook sytax is correct and renders correctly in html and epub and mobi formats. The pdf output almost renders the tables correctly apart from a few slight problems. Yelp (3.26.0) is supposed to be able to view docbook format documents and/or the html documents. It doesn't read the html documents at all (gives a URL can't be shown error) and doesn't render the tables when started with the docbook format.  It may be that it doesn't find the gnc- docbookx.dtd.
 I was unable to view the docbook using the XMLMInd editor as a viewer either. It may also need links to the xsl stylesheets Not sure how we can fix it unless we can get the Gnome developers to fix yelp, It seems silly to restrict the other document formats for the sake of complying with  a poor viewer.  yelp's limitations seem to get a fair bit of airplay on various forums. I can raise the problem on the Yelp bugs list and see if that raises any response but it won't help in the short term.

David

help/C/Help_ch_Transactions.xml
jralls: @DaveC49 A(dd) is for new transactions that don't have a match in the book, I don't think they're what we're concerned with here. 
The R checkbox on the importer marks a match as cleared, unchecking it neither adds a transaction nor changes the existing one, so I think that "C" is still correct for that column.
As for whether U+C should unreconcile the split, that seems to be a bit controversial depending on what gets changed. If the amount changes then it's obvious that the split should be unreconciled, but @gjanssens has gotten a lot of flack for his change to the register logic that unreconciles for any change at all to a reconciled split or to its parent transaction.

Changing the headings might be a little confusing to some existing users, but fixing them will be less confusing to new ones. It also changes translatable strings. Perhaps it would be better to put that part in master.

help/C/Help_ch_Transactions.xml
jralls: Might the problem with Yelp be due to a DocBook version difference?

help/C/Help_ch_Transactions.xml
DaveC49: @Bob-IT Thanks for that- solved the problem.

help/C/Help_ch_Transactions.xml
DaveC49: @jralls thanks John. I'll create a bug for it and change the U+R to U+C and R to C in the code. If Geert has copped flack for having it change the reconcile status, then that is better left in the user's hands. I'll do that separately from the other issues Frank raised above. Most of these have already been done in a couple of commits I haven't pushed through yet.

help/C/Help_ch_Transactions.xml
fellen: > Might the problem with Yelp be due to a DocBook version difference?

In /usr/share/yelp/dtd/catalog they have db 4.1.2 - 4.5 dtd's

DaveC49: I can't find any reference to what version of DocBooks yelp supports but it seems to be both 4.5 and 5.0. Nested tables has been supported since v4.4 in DocBooks.

DaveC49: I can't find any reference to what version of DocBooks yelp supports but it seems to be both 4.5 and 5.0. Nested tables has been supported since v4.4 in DocBooks.

DaveC49: I don't really know enough about how docbooks works. I'll play with a dummy table in a dummy document and see what I can make work in yelp. I've put a question up about yelp and docbook on an ubuntu forum to see what comes back and I'll raise an issue about it on the main yelp gitlab bug reporting. My guess is they have only implemented a subset of the docbook capability sufficient for writing help type files. One of the reasons for the shift to mallard in yelp seems to be it is better suited for constructing help which is task oriented and it doesn't tag everything so it is more tolerant.

fellen: The "normal" db2html processing of `<entrytbl>` is done in https://github.com/Gnucash/gnucash-docs/blob/maint/xsl/1.79.2/html/table.xsl#L696, but yelp and khelpcenter use their own xsl's. So I think, you should file a bug against yelp.

jralls: Note that we don't generate those stylesheets, they come from [DocBook's repository](https://github.com/docbook/xslt10-stylesheets/).

DaveC49: @jralls @fellen I will raise a bug report with yelp. Can't really see why they would use their own xsl's when DocBook make them available from their repository.  I have reworked the table so that it is at least presentable in yelp but so far it is problematical in HTML now.  I have set myself up properly to build out of source with a short script to run after I have pulled from GnuCash/gnucash-docs on github.

PR #126, opened 2019-07-31
Import Matcher docs Minor typos colorfixes-Simplify Table format

DaveC49:
@fellen, @jralls
John, Frank This commit is a further change to the previous documentation update on the import matcher. It specifies the actual default colors used  in gnucash-fallback_310.cssfor display and includes a simplified form of the table in the previous update which no longer has nested tables. This one comes up in HTML in Chrome and firefox, pdf in full color as intended and in yelp, epub and mobi in black and white with the color names in the table (I contemplated putting the hex codes for the colors in but if someone realy needs to know they can look up the CSS file which I referenced in the note about possible displayed color changes. AFAIK at present changes in the CSS would be on class names as I don't remember seeing any specific Id's on the elements in the Glade files which could be referenced.
 I haven't changed the U+R and R to U+C and C  with corresponding changes in the text in this set of changes - will do that in another pull request in a new branch. 
I am planning on setting up some dummy import files so I can create a screen shot of actual import situations with the matcher where there are matched transactions, transactions which have matched the transfer account tfor import, some to import without assigned transfer accounts, some matched for updating and then a second screen shot with all transactions ready for import to illustrate the situations in the table with an actual example.

### DIFF
--- a/help/C/Help_ch_Transactions.xml
+++ b/help/C/Help_ch_Transactions.xml
@@ -1140,7 +1140,7 @@ Author:
           </tbody>
         </tgroup>
       </table>
-      <note><para>The colors illustrated here are the defaults for GnuCash specified in gnucash-fallback-310.css. You may experience different colors if the gtk-3.0 CSS files have been modified</para>
+      <note><para>The colors illustrated here are the defaults for GnuCash specified in <ulink url="https://github.com/Gnucash/gnucash/blob/maint/gnucash/gnucash-fallback-310.css">gnucash-fallback-310.css</ulink>. You may experience different colors if the gtk-3.0 CSS files have been modified</para>
       </note>
       <para>Where a row has been matched to an existing transaction or is to update an existing transaction or a transfer account 
          has been matched and supplied, double clicking on the row will bring up a dialog which displays detail of the imported 
--- a/help/C/Help_ch_Transactions.xml
+++ b/help/C/Help_ch_Transactions.xml
@@ -1067,102 +1067,87 @@ Author:
     </sect2>
     <sect2 id="trans-import-matcher">
       <title>Import Matcher</title>
-      <para>The Import Matcher uses a Bayesian approach to assign a transfer account, when it is not specified in 
-          the imported data, to each imported transaction based on the previous import history of the import account.
-          It also attempts to match the transactions being imported to any existing transactions already recorded
-          based on the date, amount and the description fields by tokenizing that information and comparing it to stored tokens for 
-          assigned transfer accounts assigned in previous imports.<!-- FIXME add any other fields used in matching -->.
+      <para>The Import Matcher checks whether an imported transaction matches an existing transaction already recorded in GnuCash 
+          within a predefined time window for the account into which the transactions are being imported. A match probablity is 
+          calculated based on the Date, Amount, Transaction ID's(if present), Check numbers (if present) and the Description
+          and split Memo fields. Exact matches will generate a high match probability and flag a transaction as matching an existing
+          transaction and therefore not to be imported. If there is a lower quality match to an existing transaction, the imported transaction
+          may be flagged for updating of non-matching information in the GnuCash record of the matching transaction. These decisions can be overridden by the user.
+      </para> 
+      <para>If no matching transaction can be identified the transaction will be flagged for import. The matcher will attempt to assign
+          a transfer account based on matches between tokenized information for previously imported transactions stored in lists associated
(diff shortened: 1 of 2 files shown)


help/C/gnucash-help.xml
```diff
@@ -310,6 +310,9 @@
 	<firstname>J. Alex</firstname> <surname>Aycinena</surname>
       </author>,
       <author>
+  <firstname>David</firstname> <surname>Cousens</surname>
```
jralls: Indent, not out-dent. ;-)

fellen: @jralls I wonder, if we should define and apply a (relaxed) style for our docbook documents, e.g. indent 2 spaces, linelenght 80

jralls:
@fellen for XML I don't really care what is the indent as long as it's the same for all elements in a particular document and child elements are indented not outdented. Github displays tabs differently from most editors so I prefer spaces to make reading easier here.

help/C/Help_ch_Transactions.xml
```diff
@@ -1067,102 +1067,87 @@ Author:
...
+            </row>
+          </tbody>
         </tgroup>
-       </table>
-       <note><para>The colors illustrated here are the defaults for GnuCash on a Linux system. If you have altered the CSS
-        or are using another OS the colors may be different.</para>
-       </note>
-       <para>Where a row has been matched to an existing transaction or is to update an exitinng transaction or a transfer account 
+      </table>
+      <note><para>The colors illustrated here are the defaults for GnuCash specified in gnucash-fallback-310.css. You may experience different colors if the gtk-3.0 CSS files have been modified</para>
```
fellen: How about adding a link https://github.com/Gnucash/gnucash/blob/maint/gnucash/gnucash-fallback-310.css?

fellen:
The commit message should be amended
> Minor typos colorfixes
> 
> Add self to contributors list  fix minor typo fixes in table. Set colors to those defined in gnucash-fallback-310.css.
> Changes to matcher table for better yelp display
> 
> modified table to improve the rendering of the table in yelp
> remove gnucash-docs-3.6.tar.gz
> 
> 
> remove gnucash-docs-3.6.tar.gz
> 
> 
> Minor typo fix to importmatcher docs
> 
> 
> SImplfy Table of matcher status indicators
> 
> Nested tables didn't render well in all formats. table is now unnested and comes up looking good in yelp
> Minor edit of simplified matcher status table

The first line should have the summary, i.e. "Improve Transactions:Import matcher"
After a dividing empty line, you can enter details, ideally one per line.
* There is no gnucash-docs-3.6.tar.gz modification in this  commit.

help/C/Help_ch_Transactions.xml
DaveC49: Done

help/C/gnucash-help.xml
DaveC49: @jralls I noticed when I opened gnucash-help.xml  to fix the above that all the other author entries showed up as being out-dented not indented. Is that worth fixing? I am using xed with an indent at 2 characters and tabs converted to spaces. If I open it in emacs they all display correctly as emacsand some are indented as tabs and some as spaces. All the <author> and </author> tags are indented with 6 spaces while most of the <firstname> tags are indented with tabs (which xed displays as 2 spaces - it doesn't change tabs to spaces on read but does during editing).  Most of the rest of the document is indented with spaces and not tab stops.

jralls: David,
I don't know xed, but in emacs you can `M-< M-> M-X untabbify` to change all of the tabs to spaces, after which it should look right everywhere. Commit as a separate "whitespace" change.

christopherlam: I'd encourage emacs 👍 which seems to do things right every time!

fellen:
I have commited a modified version of this PR .

commit 357a5ff/c252bdb: I rewrote the commit message to get rid of the non existant tgz notes
commit e0b3f6c/7871715: no changes

New: 
commit 735e40a IMHO the table looks clearer with vspans, but if you dislike it, we can revert them. Because "assign a transfer account" boosts productivity, it is a tip.
commit 8c0bbcc: Housekeeping of Authors section replacing the other `<TAB>`s.

Note:
In theory we should not use `<?db* ...>` in docbook source files, but in related XSLT files, but for now it might be overhead.

@DaveC49 If you have no objections to my modifications, we can close this PR

fellen:
Closing as no objections appeared.
Thanks!

PR #127, opened 2019-09-11
Help manual Importing Transactions Bug 797337

DaveC49:
Series of updates to the Help manual Importing Transactions section to change the orientation more towards a description of the interface functionality in the help manual . Restructuring and adding screenshots to Importing QIF, Importing CSV, Importing OFX, Import Matcher. Added new sections on the Match Picker and Import Map Editor  and a new section re Exporting Transactions.


help/C/Help_ch_Transactions.xml
```diff
@@ -891,6 +891,12 @@ Author:
           </para></listitem>
           <listitem><para><emphasis role="bold"><link linkend="trans-import-csv">CSV</link></emphasis> 
                (.csv)  Comma Separated Values;</para></listitem>
+          <listitem><para><emphasis role="bold"><link linkend="trans-import-MT940-MT942-DTAUS">MT940</link></emphasis> 
+               (.STA)  SWIFT MT940 Customer Statement Message;</para></listitem>
```
fellen: 1. While the 3 first formats can hold transactions for several accounts in one file, this 3 usually hold transactions for one bank account (or CC).
1.  From where do you have this and the next 2 filename extensions? AFAIK, there are none defined.
1. The official description is "end-of-day bank account statement". [https://en.wikipedia.org/wiki/MT940] In contrast 942 is an intra-day statement.

help/C/Help_ch_Transactions.xml
```diff
@@ -891,6 +891,12 @@ Author:
...
+          <listitem><para><emphasis role="bold"><link linkend="trans-import-MT940-MT942-DTAUS">MT942</link></emphasis> 
+               (.txt  SWIFT MT942 Interim Transaction Report;</para></listitem>
+          <listitem><para><emphasis role="bold"><link linkend="trans-import-MT940-MT942-DTAUS">DTAUS</link></emphasis> 
+               (.DTAUS)  Banking data Exchange File;</para></listitem>
```
fellen: The name is derived from german "Datenträgeraustauschverfahren". [https://de.wikipedia.org/wiki/Datentr%C3%A4geraustauschverfahren]. First used 1976 with 9-track magnetic tapes, later other media like floppy. Bank support ended 1. Februar 2016.

I am not  sure: should we add this infos in the chapter or the glossary?

help/C/Help_ch_Transactions.xml
```diff
@@ -899,180 +905,543 @@ Author:
...
       <sect3>
-        <title>Load all of the QIF files containing data you wish to import</title>
-        <para>To do this, select
+        <title>To import QIF files:</title>
+        <para>Select
           <menuchoice>
             <guimenu>File</guimenu>
             <guisubmenu>Import</guisubmenu>
-            <guimenuitem>Import QIF...</guimenuitem>
+            <guimenuitem>Import QIF</guimenuitem>
```
fellen: It should be the other way: The error is in the program. This menu entries should show a hellipsis, because a dialog follows in contrast to a command, which is immediately executed like File->Save.

help/C/Help_ch_Transactions.xml
```diff
@@ -899,180 +905,543 @@ Author:
...
+          <imageobject role="html">
+            <imagedata fileref="figures/Help_Import_QIF_Druid.png" format="PNG" srccredit="David Cousens" width="510"></imagedata>
+          </imageobject>
+          <imageobject role="fo">
+                      <imagedata fileref="figures/Help_Import_QIF_Druid.png" format="PNG" srccredit="David Cousens"></imagedata>
+          </imageobject>
+          <textobject>
+            <phrase>Import QIF files dialog window.</phrase>
+          </textobject>
+          <caption><emphasis role="bold">Import QIF files dialog window.</emphasis></caption>
```
fellen: No need for emphasis here. Instead use:
```
<figure pgwide="1">
    <title>Import QIF files dialog window</title>
    <screenshot>
    : 
   </screenshot>
</figure>
```
See https://wiki.gnucash.org/wiki/Documentation_Update_Instructions#Screenshots_and_Images

help/C/Help_ch_Transactions.xml
```diff
@@ -899,180 +905,543 @@ Author:
...
-              displays the percentage finished. When the import process is complete, GnuCash will return you 
+        <screenshot id="ImportQIFDruid">
+        <mediaobject>
+          <imageobject role="html">
+            <imagedata fileref="figures/Help_Import_QIF_Druid.png" format="PNG" srccredit="David Cousens" width="510"></imagedata>
+          </imageobject>
+          <imageobject role="fo">
+                      <imagedata fileref="figures/Help_Import_QIF_Druid.png" format="PNG" srccredit="David Cousens"></imagedata>
+          </imageobject>
+          <textobject>
```
fellen: Textobjects should be descriptions of the image, so that screenreaders can tell blind people something about the image. Repeating the title makes no sense. If you really want a text object fill it with the strings from the sources.

help/C/Help_ch_Transactions.xml
DaveC49: Frank, 
I can add a note to the effect that the MT940 is an end of day and MT942 is an intra day statement. The .STA extension came form a Polish bank's description of the MT940 format but it is probably just their practice. I also used http://www.societegenerale.rs/fileadmin/template/main/pdf/SGS%20MT940.pdf when I was writing it for info. Found it very difficult to locate official SWIFT documentation. noreference to MT940 or MT942 exists on their website possible because ISO 20022 has superseded them for interbank use. I presume the ISO 20022 is a front end to encrypted link for the exchange. 

If the use of a specific extension is not mandatory or even  common I will drop it . I lumped the three formats together primarily because there is little setup involved in importing them apart from loading the specific  files as the internal format of the file is quite strict. I had a couple of sample DTAUS  MT940 and MT942 files to play with that I downloaded from a couple of European bank sites.

help/C/Help_ch_Transactions.xml
```diff
@@ -899,180 +905,543 @@ Author:
...
+            <phrase>Import QIF files dialog window.</phrase>
+          </textobject>
+          <caption><emphasis role="bold">Import QIF files dialog window.</emphasis></caption>
+        </mediaobject>
+        </screenshot>
+        <para>On most panes the dialog has three buttons on the right for navigating through the dialog steps shown in the left pane of the dialog:
+          <itemizedlist>
+            <listitem>
+              <guibutton><emphasis role="bold">Cancel</emphasis></guibutton>
```
fellen: Nitpick: If you think, we really need the emphasis, I would prefer `<emphasis role="bold"><guibutton>Cancel</guibutton></emphasis>` because the emphasis is not part of the gui element. Search-and-replace operations will be much easier. That has several advantages. 

1. Imagine GTK renames `<guibutton>Cancel</guibutton>` to `<guibutton>Abort</guibutton>`
2. There is https://bugs.gnucash.org/show_bug.cgi?id=633673. We probably want convert it then from
```
<itemizedlist>
   <listitem>
      <emphasis role="bold"><guibutton>Cancel</guibutton></emphasis>
      <para>Exit the dialog;</para>
   </listitem>...
```
to
```
<variablelist>
   <varlistentry>
      <term><guibutton>Cancel</guibutton></term>
      <listitem><para>Exit the dialog;</para></listitem>
   </varlistentry> ...
```

help/C/Help_ch_Transactions.xml
```diff
@@ -899,180 +905,543 @@ Author:
...
+    
     <sect2 id="trans-import-ofx">
       <title>Import OFX/QFX</title>
-      <para>This opens a file selection dialog. Navigate to the file you wish to import, select a file with 
-        the appropriate extension (.ofx  or .qfx), then press the <guibutton>Import</guibutton> button button.</para>
-      <para>Gnucash opens an Account Selection dialog to select an account in your CoA corresponding to data source. Select
-        the appropriate account from the account tree and press the <guibutton>OK</guibutton> button. On subsequent import of files from the 
-        same source (identified by tags in the file), the source is remembered and the account selection dialog is not
-        displayed.</para>
+      <para><guimenu>File</guimenu>-><guisubmenu>Import</guisubmenu>-><guimenuitem>Import OFX/QFX transactions</guimenuitem>
```
fellen: Should we explain OFX and QFX? Again, here or in the glossary.

help/C/Help_ch_Transactions.xml
```diff
@@ -899,180 +905,543 @@ Author:
...
       </para>      
     </sect2>
+    
     <sect2 id="trans-import-csv">
       <title>Import CSV</title>
-      <para>Clicking on <guimenuitem>Import CSV...</guimenuitem> in the Import menu will bring up the Import Assistant dialog. The first step brings up a file selection 
-            dialog. Navigate to the location where the file you wish to import is located and select the file to import then click the 
-            <guibutton>OK</guibutton> button.
+      <para>Clicking on <emphasis role="bold"><guimenu>File</guimenu>-><guisubmenu>Import</guisubmenu>-><guimenuitem>Import Transactions from CSV</guimenuitem></emphasis>
+        will bring up the CSV Transaction Import Assistant dialog illustrated below.
```
fellen: Mention lowest level of exchange format? All others are structured financial data. CSV is a plain table.

help/C/Help_ch_Transactions.xml
```diff
@@ -899,180 +905,543 @@ Author:
...
+                  </textobject>
+                  <caption><emphasis role="bold">Import Preview Pane with settings to import the data shown.</emphasis></caption>
+                </mediaobject>
+              </screenshot> 
+              <itemizedlist>
+                <listitem>
+                  <emphasis role="bold"><guilabel>Load and Save Settings</guilabel></emphasis>
+                  <para id="trans-import-csv-save">Once you have set all the import parameters, you can save these settings by typing
+                    in a setting a unique name in the <guilabel>Load and Save Settings Entry</guilabel> combo box and clicking the
+                    <inlinemediaobject>
```
fellen: You know the buttons depend on your desktops selected appearance aka "theme"?

help/C/Help_ch_Transactions.xml
```diff
@@ -1105,116 +1502,544 @@ Author:
...
+        <itemizedlist>
+          <listitem> change the import action by selecting another checkbox (or none); and/or</listitem>
+          <listitem>manually assign a different transfer account;</listitem>
+        </itemizedlist>
+        after reviewing the matched transactions and accounts and the automatically selected actions.
       </para>
- 
+      <para>Multi row selection is enabled for the list of downloaded transactions. The following operations can be performed on the list of downloaded transactions:</para>
+      <itemizedlist>
+        <listitem><emphasis role="bold"><phrase>Single<mousebutton>Left Click</mousebutton> on a row</phrase></emphasis>
```
fellen: How about 
`<keycombo action="click"><mousebutton>Button1</mousebutton></keycombo>`,
`<keycombo action="double-click"><mousebutton>Button1</mousebutton></keycombo>`, ...
See examples in docbooks keycombo page.

help/C/Help_ch_Transactions.xml
DaveC49: It would be nice to link to external definitions but the volatility of that is not controllable. My personal preference is for explaining formats where they are used unless we can link directly into the glossary. What would be nice  would be to have sections from a glossary appear as popups when you hover on them, at least in the html if not the other formats.  I was not familiar with glossries in DocBook but it appears you can have them linked from the text.  I will have to experiment. Is there currently a glossary implemented in the documentation? Would you envisage the glossary to be attached to a chapter or to be documentation wide? I gather both are possible http://www.sagehill.net/docbookxsl/GlossDatabase.html. AFAI can see at the moment we don't have a general glossary but it would obviously be useful when it comes to terms like "split" which have a GnuCash specific meaning as well as the many acronyms which come up which could all be defined in a document.  It would appear that you have to setup an XML  catalog to do the most general approach for a multidocument gloassary..

help/C/Help_ch_Transactions.xml
DaveC49: Will do

help/C/Help_ch_Transactions.xml
DaveC49: That makes life a bit more difficult.  OK will stick to a generic label description.  I guess somewhere we will need a disclaimer pointing out that gui element appearance in screenshots may  be OS and theme dependant and should not be relied upon.

help/C/Help_ch_Transactions.xml
fellen: I would consider wikipedia as trustworthy and stable. I defined several base URLs in docbook/gnc-docbookx.dtd.
To link https://en.wikipedia.org/wiki/QFX_(file_format), you can use
`<ulink url="&url-wp-en;QFX_(file_format)" />`

@sunfish62 created a glossary in guide. IMHO as a referential element it would belong into help, but until we fix linking between the docs you could use 
`<ulink url="&url-docs-build-C;gnucash-guide/gnc-gloss.html" />`

help/C/Help_ch_Transactions.xml
fellen: From https://travis-ci.com/Gnucash/gnucash-docs/builds/127120215:
Help_ch_Transactions.xml:937: element caption: validity error : Element emphasis is not declared in caption list of possible children

help/C/Help_ch_Transactions.xml
DaveC49: Perhaps setting up a gnucash XML catalog file in the docbook directory of the sources is the way to make the @sunfish62 glossary available to all documents which would be good. Then if entries have to be edited it only has to happen in one place. Once it exists it would require creation of an environment variable XML_CATALOG_FILES, if it doesn't already exist and appending the URL to it if it does, so xsltproc can locate it during processing. I assuming in ignorance that  we are using xsltproc and not one of the alternatives. That way the glossary is independent of both the guide and help and available in both.  I will put all the other corrections in first  then come back and experiment with making the glossary more general. @sunfish62 may have some suggestions re this as well as he more than likely had this in mind when creating the glossary. The glossary already  has entries for the various import export formats already except for DTAUS and MT942. I noticed David T used link ids so doing a global replace  of <text>with <glossterm OFX linkend="gnc_gl_ofx"><text></glossterm>  will fix allow using the glossary in the help file.

help/C/Help_ch_Transactions.xml
DaveC49: Frank  I'll remove the text objects at least for the moment until i can conceive of a way to describe a screen to a blind person. and fix up the figures

help/C/Help_ch_Transactions.xml
DaveC49: Point taken on the <emhasis role="bold". outside the <guibutton> type tags. Will find and change.

I agree with Christian Marchi's comment re the <variablelist>. I tried using it in a number of places and found the formatting extremely confusing and the meaning less clear.

help/C/Help_ch_Transactions.xml
DaveC49: Added a bit to the introductory para in Import Formats pointing this out.

help/C/Help_ch_Transactions.xml
fellen: To get an impression of the text objects, you can open the html version with an textmode browser like links and it's derivates (lynx, elinks, ...):
`links https://code.gnucash.org/docs/C/gnucash-help/trans-import.html`

DaveC49:
Frank'

I've ammended as per your comments above with the exception of the glossary. I raised a new bug 797426 to modify the glossary so it can be used with both the help manual and the guide and I will link into that once i have experimented with getting it up. In David T's inital setup it displays the glossterm in a popup but it would be much nicer if ti popped up the glossdef . Both can have id attributes so it may be possible to use links to the glossdef part of a glossentry in the linkend for the inline glossterm in the main text. I will have to experiment a bit with that. If it doesn't work out of the box it may be possible to modify the html processing file.
make check
make html 
make pdf
make epub
make mobi
all run. I have checked the html and pdf out in detail

fellen:
It looks nice. I found a few typos and had had a few minor changes/questions:

- the "UTF-8 variant for your locale" do not exist. Only the old ISO encodings had them.
- ~+_42 ??? was `"&plusmn;"`: "&plusmn;" intended? [iso-num.html in TDG]
- A match probablity is calculated ... Amount (+/- Edit->Preferences->Online Banking->Commercial ATM fees treshold)? @Gnucash/gnucash-developers 
- <keycombo action="double-click"> has not the desired effect, if it is no combo. Open for better ideas.
- I renamed your table to "Import Status of Transaction" because we have also Reconciliation status in other context and probably more. 
- ":" in a new line produces a space before it.
- I found the list of lists in "The generic import transaction matcher" confusing. So for now I did some experiments with list titles. Can you review the section?
- I stepped up the tip to important. We write already important in the text.-
- "(see below)" is discouraged in hypertext, use links.
- Figure titles should be unique, but short. There is a List of Figures in most output formats Full sentences belong in a caption section at the end of the mediaobject.
- I added a few id's and links. While I reviewed them I found in referenced titles we should replace gui... tags by simple quote tags. Else they looked ugly and very different between yelp and HTML. Please review them, also for consistency.
- Can you verify, you editor 
-- inserts 2 spaces, if you press [TAB] and 
-- removes trailing whitespace on saving?

PR #130, opened 2020-01-30
Remove redundant leading words in the admonition blocks

xuxinhang:
... and other places such as "NOTE" or "Warning".


xuxinhang:
Thanks, it is a good idea to expand some "small" and "inline" admonitions into an admonition block. I will notice it!

PR #131, opened 2020-04-22
Note: csv header lines are not supported

flywire:


guide/C/ch_importing.xml
```diff
@@ -622,7 +622,7 @@
         <listitem><para><emphasis role="bold">Encoding</emphasis> This is usually the UTF-8 variant for your locale;</para></listitem>
         <listitem><para><emphasis role="bold">Date Format</emphasis> This does not default to the Locale setting so check it matches the data you are importing;</para></listitem>
         <listitem><para><emphasis role="bold">Currency Format</emphasis>;</para></listitem>
-        <listitem><para><emphasis role="bold">Leading Lines to Skip</emphasis>Note: csv header lines are not supported and must be skipped;</para></listitem>
+        <listitem><para><emphasis role="bold">Leading Lines to Skip</emphasis>Note: A csv header line with field names is not supported and must be skipped;</para></listitem>
```
fellen, line 625: I would prefer `Example:` over `Note:`

Background we are in the process of replacing `Note:` by the admonition
```
<note>
  <para>...
  </para>
</note>

```

flywire: Thank you, I'll keep that in mind for the [Importing Data Tutorial](https://chat.stackexchange.com/transcript/message/54159828#54159828) which would have been much easier and quicker in [MkDocs](https://www.mkdocs.org/).

Intending to keep formatting consistent with the previous few lines. Example is wrong in this context so is it reasonable to just delete the word _Note:_? Would the term _Fieldname Header_ be clear enough? I'd be happy to be directed to a better place to discuss.

gjanssens: Actually both are too limited. A header line with field names is only one example of leading lines. I have also seen csv files from banks that start with a number of summary lines (statement date, statement number, start balance, bank account id,...) and only then continue with the actual csv data. In the face of such a csv file only mentioning fieldname header or variants is not helpful.

Also I think "not supported" is a confusing term. It implicitly suggests the leading lines contain importable data which gnucash is not capable of doing. The reality is there is too much variation in that additional data it's impossible to write an efficient interpreter for it. I'd prefer some other way of clarifying this option.

Perhaps something like
> Use this to tell the importer to ignore non-transaction lines preceding the actual data. This can include a field name header or an introduction section in the import file.

flywire: > A header line with field names is only one example of leading lines.

That was the only use case I was intending to describe, I'd expect it's the most common type of csv header and certainly one that people [me] might expect would just be used like other software if the field names match.

fellen:
A few notes about the current status:
1. @DaveC49 is currently still working on the improvement of the import theme. You should coordinate your work with him.
1. The current chapter in guide is an outdated copy of the corresponding section in help/transactions. We had a collision: while @gjanssens fixed some isssues of @sunfish62's "Move Import vom Help to Guide", I applied @DaveC49's "Improve Import".
My suggestion: Let us finish at first the work in help (technical details) before we rework the import section of the Guide (How to, ...)
FYI the current section in help
```
<listitem>
  <para><emphasis role="bold"><guilabel>Leading Lines to Skip</guilabel></emphasis>:
    Select the number of header lines to skip in the imported file;</para>
</listitem>
<listitem>
  <para><emphasis role="bold"><guilabel>Trailing Lines to Skip</guilabel></emphasis>:
    Select the number of trailing lines to skip in the import file;</para>
</listitem>
```
IIRC this are the current tooltips of the program (our minimum requirement for help).
Easiest would be to add a common footnote: While we only want to import transactions, some banks send additional informations like a fieldname header or account status information as header or footer. The alternatives would be 
1. a common intro.
2. separate examples fieldname for the header, account summary for the footer.

fellen:
As no further commit or comment appeared, i applied the suggestion from my last comment.

Thanks!

PR #138, opened 2020-05-09
Modifications to Guide Ch12 to include multi-currency transactions us…

DaveC49:
…ing trading accounts.

Split the multi-currency transaction after the initial introduction into two sections. The first is the original dealing with the treatment of multicurrency transactions using manual recording of the gains and losses.  The second new section deals with gains and loss treatment using trading accounts. basic introduction to trading account concepts added and an example of a transfer of funds between accounts in different currencies is added (incomplete)
Complete currency trading example

typo and tag fixes

typo and tag fixes

tag and typo fixes

type and tag fixes

Updates to screenshots

Fix up screenshots

fix tags

changes after proof reading

tag mismatch fix

Proof reading edits

typo fix


guide/C/ch_currency.xml
```diff
@@ -34,793 +35,1132 @@
...
+      <sect3 id="currency_trading_account_setup">
+        <title>Setup of Accounts for Trading Accounts</title>
+        <para>The same account structure useds in the manual recording of
+          currency transactions in the previous section is required. The
+          initial account configuration for Assets and Equity accounts are
+          as follows:</para>
+        <screenshot id="currency__trading_acct_setup">
+          <mediaobject>
+            <imageobject role="html">
+              <imagedata fileref="figures/currency_trading_acct_setup.png"
```
fellen, line 903: 2 Issues:
* The important part is not displayed: The trading accounts. You should resize the window before capturing the screenshot again.
* This is IMHO already wrong in 12.3:
Equity and its subaccounts like `Opening Balance` should be only in your local currency.
We should not give the users bad examples.
Or missed I a reference In the later parts of this chapter? I suspect the original author tried to explain: As long as there is no exchange rate, the value `Total (<local currency>)` of foreign currencies is displayed as 0. We should try to find an better example to demonstrate it in the common part.

Then the price would not entered manually, but by the exchange dialog while creating the opening balance transactions.
NB.: Also the exchange dialog should usually "Pieces" in both currencies instead of an inaccurate exchange rate.

guide/C/ch_currency.xml
```diff
@@ -34,793 +35,1132 @@
...
-        <imageobject>
-          <imagedata fileref="figures/currency_main1.png" format="PNG"
-                     srccredit="Geert Janssens" />
-        </imageobject>
-
-        <textobject>
-          <phrase>Initial multi currency Account Bank Setup</phrase>
-        </textobject>
+    <para>You can either deal with multiple currency transaction with a manually entered transactions
+    to record currency gains and losses as described in the first section of this chapter or you can
```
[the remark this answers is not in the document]
DaveC49: Agreed better to link

jralls:
Does it really matter if some transactions involve trading accounts and some not? AFAICT they change the UI a little but don't really do anything else.

guide/C/ch_currency.xml
jralls: > Equity and its subaccounts like Opening Balance should be only in your local currency.

Then you'd only be able to record income and expense in local currency so the exchanges wouldn't be modeled in the books, hiding currency exchange gains and losses.

DaveC49: Frank,

The trading accounts did not exist at that point as they are not created until you add a transaction between currencies.

Re 12.3  There is an assumtion here that a user does not have foreign currency accounts in place with existing balances when a book is created or for those who start a new book for each new year for example. Is that a necessary restriction to place on GnuCash?

 I could remove any reference to how the accounts reached there initial state and perhaps make the example a bit more realistic for someone actually involved in currency trading. I don't think most would regard an 0.82c loss with $60.00 in costs a viable business transaction. Having explored how trading accounts work I can probably come up with a more realistic example now.

guide/C/ch_currency.xml
```diff
@@ -34,793 +35,1132 @@
...
-      <mediaobject>
-        <imageobject>
-          <imagedata fileref="figures/currency_main1.png" format="PNG"
-                     srccredit="Geert Janssens" />
-        </imageobject>
-
-        <textobject>
-          <phrase>Initial multi currency Account Bank Setup</phrase>
-        </textobject>
+    <para>You can either deal with multiple currency transaction with a manually entered transactions
```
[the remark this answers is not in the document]
DaveC49: Agreed much less clumsy construction.

guide/C/ch_currency.xml
DaveC49: This was really a first pass attempt constructed while exploring trading accounts to better understand them myself. The treatment in the 12.2 part could probably be revised in the light of Selinger's article to use manual recording using the trading account concept or perhaps dropped completely as Frank suggested. 

I haven't yet checked out what happens for example when a currency price change occurs and is recorded in the price database independent of a currency transaction between the currencies. This would affect the conversion to the Report currency in the Accounts tab but what effect it has on reports for example i.e. how do trading losses and gains in the Balance sheet when they are unrealized gains as opposed to the realized gains here. I flagged it as something to explore further while doing this exercise.

I followed the original author in setting up the accounts. How the accounts are setup is irrelevant to the operation of the trading accounts themselves!

fellen:
> Does it really matter if some transactions involve trading accounts and some not? AFAICT they change the UI a little but don't really do anything else.

As a rule of thumb:

- that without trading accounts can lead to an Imbalance,

-  that with trading accounts will instead lead to pending proftis or losses.

jralls:
I don't think that's true. Trading accounts make the trading gains and losses more visible but IIRC don't automatically book them. One must still create the realized gain/loss splits to keep the book balanced.

guide/C/ch_currency.xml
fellen: > > Equity and its subaccounts like Opening Balance should be only in your local currency.
> 
> Then you'd only be able to record income and expense in local currency so the exchanges wouldn't be modeled in the books, hiding currency exchange gains and losses.

Hm, `local currency` is ambiguous. I meant in the book currency. For the german rules you could talk with CPA assistant "Frakturfreak" on IRC.

fellen:
> I don't think that's true. Trading accounts make the trading gains and losses more visible but IIRC don't automatically book them. One must still create the realized gain/loss splits to keep the book balanced.

Right, it seems Chris did fix that.

christopherlam:
Huh, me? Did not.

guide/C/ch_currency.xml
DaveC49: Frank,
See the problem. Under the Handelgesetzbuch, gains and losses from currency transactions are recognized in profit and loss as part of other operating Income or Other Operating Expenses in an income statement rather than as line items in the Balance Sheet. 

however, the IFRS also treats gains and losses from foreign currency transactions as part of profit and loss as does the FASB in the US AFAIK ( only a brief scan of the FASB rules) and the AASB  in Australia. Germany however does not recognize "comprehensive income" which can include unrealized gains and losses as well as realized gains and losses (depends primarily on whether they want to tax you on them or not - usually not as it is against basic tax principles) in some jurisdictions.  I haven't yet ascertained how trading accounts in GnuCash are affected by unrealized gains (to be investigated) and whether the reported gains and losses are totally realized or a combination of realized and unrealized losses.

 Normally the transactions to the trading accounts occur at the exchange rate current at the time of the transactions. What is presented in the Accounts tab in the Report currency column will be affected by any changes to the currency's exchange rate recorded in the price database, but Reports in most jurisdictions for foreign currencies seem to be based on the exchange rates at the report date

I must admit I was somewhat confused that trading gains and losses were reported in the Balance Sheet in the reports.  I think a lot of these issues relate more to presentation in reports than the underlying recording in GnuCash but correctly translating what is recorded in GnuCash to the reports.

David

jralls: > I must admit I was somewhat confused that trading gains and losses were reported in the Balance Sheet in the reports. I think a lot of these issues relate more to presentation in reports than the underlying recording in GnuCash but correctly translating what is recorded in GnuCash to the reports.

It falls out from the use of pricedb prices (either nearest in time or latest) to calculate the value of the other-commodity accounts in the report currency. Since those values are different from the transaction values it creates an imbalance that must be accounted for somewhere. The report author chose (reasonably, IMO) to create an "unrealized gains" line for it.

fellen: It seems, to fulfill the 

- **low value** principle (traditional HGB), german and probably others users must use a _cost based_ price source (with the downside no longer having a foot note with the pending result).

- **fair value** principle (IFRS) users have to choose one of the _time based_ sources.

This still needs to be incorporated in tooltips and docs - after verification.

guide/C/ch_currency.xml
```diff
@@ -34,793 +35,1132 @@
...
+    <title>Automatically Recording Currency Transactions using Trading Accounts</title>
+
+    <para>Trading accounts were introduced as a feature from  &app; version 2.3.14. The
+    conceptual basis for trading accounts is described in
+    <ulink type="http" url="http://www.mathstat.dal.ca/~selinger/accounting/tutorial.html">
+    this tutorial by Peter Selinger</ulink>. A second tutorial
+    <ulink type="http" url="http://www.mathstat.dal.ca/~selinger/accounting/gnucash.html">
+    by Peter Selinger</ulink> describes the manual creation and use of trading accounts in &app;
+    prior to version 2.3.14.</para>
+
```
fellen, line 840: 1. I think this section belongs in the intro part (before currency_manual), so the user can decide where to continue.

1. I am missing a warning about potential pitfalls as decribed by Selinger.

guide/C/ch_currency.xml
```diff
@@ -34,793 +35,1132 @@
...
+    accounts are not enabled by default in GnuCash and must be explicitly enabled as described in that
+    section.</para>
 
-        <caption><para>Initial set up of 3 bank accounts with different
-        currencies. </para></caption>
-      </mediaobject>
-    </screenshot>
-    <para>As you see, the overall balances do not yet reflect any value for EUR 
-    or HKD holdings. Adding currency exchange rates will fix this.</para> 
+    <para>The following sections are common to both manually recording gains
```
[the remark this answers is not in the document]
fellen: I have put this paragraph in 

```
<!-- TODO: add sections
TODO -->
```

for now.

fellen:
Committed, thanks.
I have marked that discussions as resolved, which went in commit 1e5191f .
But there are still enough ToDo's mentioned here for the next PR. ;-)

PR #139, opened 2020-06-09
Update fq-sources.html: asx source is working again

goodvibes2:
Also explain why you need F::Q 1.41


fellen:
I am currently patching translations ...

fellen:
Applied, thanks!

Just a note: Ideally mention the related bug like I did in the next commit.

PR #144, opened 2021-04-09
Transfer all missing help files to help/de.

CWehli:
These chapters are all untranslated so far.  A note that the German translation is not yet complete is added to all
files at the beginning of the section.


fellen:
I miss a preparing step:
Thanks @sunfish "Welcome to Gnucash" is now in C/guide §2.3.1 and still missing in de/guide.
There are probably more parts, which should be saved/moved?
The changes happened in commit 755d33c - commit 7a37189

fellen:
OTOH you are on the right track: after basing PR #142 on this, make check does no longer fail.

CWehli:
Indeed. Since the two parts "Guide" and "Help" are considered separately in all languages, there will always be differences between the translations.

Since this section you mentioned is not part of this PR, the "older" translations would have to be checked against the originals in help/C.

CWehli:
And the plan of this PR was to enable the error-free installation of the sections in help/de and to see what still needs to be translated afterwards.

CWehli:
@fellen: The formatting in gnucash-help.xml in section <revhistory> <revision> <revnumber> etc., from line 63 onwards, is not resolved correctly by xmlformat.pl. Maybe you have an idea to improve this?

fellen:
> @fellen: The formatting in gnucash-help.xml in section etc., from line 63 onwards, is not resolved correctly by xmlformat.pl. Maybe you have an idea to improve this?

xmlformat.conf:30
```
# No processing:
abbrev address literallayout programlisting screen synopsis
  format = verbatim
```
because of https://tdg.docbook.org/tdg/4.5/address.html
```
Processing expectations

Formatted as a displayed block. This element is displayed “verbatim”; whitespace and linebreaks within this element are significant.
```
resulting no formatting by xmlformat.

After manually adjustment it syncs again with C.

I removed already superflous `<address>` tags by reordering, but to attach an email to the publisher I saw no other possibility.

PR #168, opened 2021-05-29
Fixes reference to Security Editor on Help Manual

igorlvicente:
Let me know if the PR requires more than just that.

### DIFF
--- a/guide/C/ch_invest.xml
+++ b/guide/C/ch_invest.xml
@@ -818,7 +818,7 @@ Income
                   <para>The <guilabel>Type:</guilabel> should already be <guilabel>NASDAQ</guilabel>, because this is what
                     was selected in the security selector, but you can change it here, including
                     adding more categories. More information about this can be found in the Help
-                    Manual in section 7.7, <quote>Security Editor</quote>.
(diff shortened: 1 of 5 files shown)


fellen:
Hi,
nice to apply it in all languages!
That will fix it for now, but inserting or removing a chapter before the the target will break it  again.
A link would then update its description.

if it were in the same document, we could use `<xref linkend="">` here.
The proper links to other documents are olinks, but we have to set them properly up in the build process. 

Until that is done, we can use a link to the document on the build server:
`<ulink url="&url-docs-build-C;help/tool-security-edit.html">content of quote</ulink>`
"C" is the language symbol for english and should be replaced
like replace
`na secção 7.7, <quote>Editor de mercadorias</quote>.`
by
`na secção <ulink url="&url-docs-build-pt;help/tool-security-edit.html">Editor de mercadorias</ulink>.`

Do you want to improve it?

fellen:
For now I have merged it – thanks!

Do have plans to improve it like i described?

igorlvicente:
Yes, I want to improve it but I don't know when I will be able to.

fellen:
Don't hurry. And ask if something is not obvisiuos. When you are ready create a new PR.

PR #185, opened 2021-07-14
Fix ch_txns.xml

usamimikamen-sk:
Replace starting balance with opening balance.
Add comma to the numbers over 1000.


fellen:
Thanks!

PR #188, opened 2021-07-14
Replace illeagal acronym tag with guilabel tag.

usamimikamen-sk:
While the word "XML" is an acronym, xml in GnuCash save dialog is not an acronym but a guimenuitem.
The same is true for sqlite3, postgres and mysql.

### DIFF
--- a/guide/C/ch_basics.xml
+++ b/guide/C/ch_basics.xml
@@ -1156,7 +1156,7 @@ Translators:
 
         <listitem>
           <para>Select the <guilabel>Data Format</guilabel> of the file you are saving from the drop down list. The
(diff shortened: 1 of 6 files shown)


fellen:
Right, the main purpose of acronym seems to be to decide if a dot finishes an abbreviation or a sentence,

Almost, guimenuitem is a leave in the menu tree. But in the program they appear as listelements in the File->Save as dialog. For now we should use guilabel like in other parts.

Another possibility might be to define them in the glossary and then use firstterm/glossterm, but I had no time to test the appearance and functionality.

usamimikamen-sk:
I'm sorry.
I made many mistakes in my repository, so I close this pull request.

PR #190, opened 2021-07-15
Replace misused acronyms with guilabels.

usamimikamen-sk:
Replace misused acronyms with guilabels.
Add spaces between words and tags to avoid word concatenation.
Change German word Strg to English word Ctrl.
Fix typos.

### DIFF
--- a/guide/C/ch_basics.xml
+++ b/guide/C/ch_basics.xml
@@ -483,7 +483,7 @@ Translators:
         in the <guilabel>Welcome to &appname;!</guilabel> menu, or when you select <menuchoice>
           <shortcut>
           <keycombo>
-            <keycap>Strg</keycap><keycap>N</keycap>
+            <keycap>Ctrl</keycap><keycap>N</keycap>
           </keycombo>
           </shortcut> <guimenu><accel>F</accel>ile</guimenu><guimenuitem><accel>N</accel>ew
           File</guimenuitem>
@@ -510,7 +510,7 @@ Translators:
             These items are explained elsewhere in the Guide, and can be changed at any time from <menuchoice>
               <shortcut>
               <keycombo>
-                <keycap>Strg</keycap><keycap>Enter</keycap>
+                <keycap>Ctrl</keycap><keycap>Enter</keycap>
               </keycombo>
               </shortcut>
               <guimenu><accel>F</accel>ile</guimenu><guimenuitem>Proper<accel>t</accel>ies</guimenuitem>
@@ -603,7 +603,7 @@ Translators:
             the process, back up to change something, or apply to confirm your choices. If you're
             creating a new book, &app; will present a dialog to select a backend and either a file
             to save it or database server, database name, and credentials to create a MySQL or
-            Postgresql database if GnuCash was built with those features and you have the necessary
+            PostgreSQL database if GnuCash was built with those features and you have the necessary
             dependencies installed.
           </para>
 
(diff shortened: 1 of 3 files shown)


guide/C/ch_basics.xml
```diff
@@ -483,7 +483,7 @@ Translators:
         in the <guilabel>Welcome to &appname;!</guilabel> menu, or when you select <menuchoice>
           <shortcut>
           <keycombo>
-            <keycap>Strg</keycap><keycap>N</keycap>
+            <keycap>Ctrl</keycap><keycap>N</keycap>
```
fellen, line 486: Right, this happened, when we backported improvements from the German translation.

Currently only translators are working on the docs.

guide/C/ch_basics.xml
```diff
@@ -603,7 +603,7 @@ Translators:
             the process, back up to change something, or apply to confirm your choices. If you're
             creating a new book, &app; will present a dialog to select a backend and either a file
             to save it or database server, database name, and credentials to create a MySQL or
-            Postgresql database if GnuCash was built with those features and you have the necessary
+            PostgreSQL database if GnuCash was built with those features and you have the necessary
```
fellen, line 606: I wonder if the 3 SQL apps should be added to the DTD with markup.

guide/C/ch_txns.xml
```diff
@@ -1029,7 +1029,7 @@
       create it for you.
     </para>
 
-    <para>In this howto, we&rsquo;ll take a monthly Internet subscription of 20 USD as example, which is taken
+    <para>In this how-to, we&rsquo;ll take a monthly Internet subscription of 20 USD as example, which is taken
```
fellen, line 1032: IMHO we should rename all `howto`s to exercise to match with `tutorial` in the doc name.

guide/C/ch_txns.xml
```diff
@@ -353,9 +353,9 @@
         <listitem>
           <para>All of the menu items have access keys defined, and these are marked by underlined characters in the
             menu names. Press <keycap>Alt</keycap> + [underlined character] to bring up the menu,
```
fellen, line 355: `[<replacable>underlined character</replacable>]` ?
Can you test which looks nicer?

fellen:
Merged, but I added a few comments for future improvements.

Thanks!

guide/C/ch_basics.xml
usamimikamen-sk: That is a difficult problem.
For example, the official name of the application is MySQL.
However, in the select box it is mysql. The same is true for PostgreSQL, SQLite3 and XML.
If the writer and translator do not misunderstand it, you can add it as ENTITY.

Currently, my priority is to fix typo that interferes with translation.

guide/C/ch_txns.xml
usamimikamen-sk: I can't understand the difference between them, because I'm not a native English user.

guide/C/ch_basics.xml
usamimikamen-sk: I think shortcuts and acceleration keys are not necessary in this guide.
Write them in the help manual instead.

guide/C/ch_txns.xml
fellen: Me too. That's why I suggest different term.
@jralls our native speaker

guide/C/ch_basics.xml
fellen: That can be resolved by defining app-foo <application>Foo</application> and
gnc-foo <userinput>foo</userinput>.

guide/C/ch_txns.xml
usamimikamen-sk: Todo.

guide/C/ch_txns.xml
jralls: I like exercise.

The usage manuals will tell you to use how-to, but they're old and I believe it's pretty common to leave out the hyphen and Apple's spell checker doesn't complain about it.

usamimikamen-sk: I can understand what you say.
However, it is important to unify the terms in order to improve the quality of software and documents.

You may change to exercise soon, but for now it's better to unify with how-to, I think.

guide/C/ch_txns.xml
usamimikamen-sk: I think the former is better, because underlined character is not keycap right now.
Forthermore, it is better to change xsl for replaceable to be parsed as italic font.
But I don't have less knowledge about xsl stylesheet.

fellen: The square brackets serve as our replacement in plain text for Docbook's keycap tag. As we are here in docbook, we should use keycap.

Many former contributors didn't know much more about docbook than "Put your text in a para tag to get a linebreak" resultng in many odd constructions.

On my machine I see it in yelp (xml), firefox (html), and pdf in italic. Where is it not?

usamimikamen-sk: I'm sorry.
They are shown in italic in my environment.
I just read a wrong file.

Thank you for your confirmation.

PR #198, opened 2021-07-20
Fix guide-ch_currency.xml

usamimikamen-sk:
Change a paragraph to itemizedlist to make translation easier.
Change a hardcoded numbered list to docbook numbered list.
Replace starting balance with opening balance.
Fix typos and minor mistakes.
Fix RewordMiles account name.
Change CoA to Chart of Accounts
Unify currency expression like 12,345 USD for foreign currencies.
Normailze account names.
Change one of the note because English sentenses are too difficult.
Change the date format like Jan. 2, 2020.
Replace Price Editor with Price Database but images are not changed.
Change +ve and -ve to positive and negative, respectively.
Fix according to the pull request comment.

### DIFF
--- a/guide/C/ch_currency.xml
+++ b/guide/C/ch_currency.xml
@@ -28,9 +28,12 @@
       example, you can have a bank account set up in Euros, and another using Hong Kong Dollars.
     </para>
 
-    <para>Some of the issues which arise when using multiple currencies are, how do you transfer funds between
-      accounts with different currencies? How do you calculate the overall value when you have mixed
-      currency accounts? How do reports deal with mixed currencies?
+    <para>Some of the issues which arise when using multiple currencies are as follows:
+    <itemizedlist>
+        <listitem><simpara>How do you transfer funds between accounts with different currencies?</simpara></listitem>
+        <listitem><simpara>How do you calculate the overall value when you have mixed currency accounts?</simpara></listitem>
+        <listitem><simpara>How do reports deal with mixed currencies?</simpara></listitem>
+      </itemizedlist>
     </para>
 
     <para>You can either deal with multiple currencies using manually entered transactions to record currency
@@ -70,7 +73,7 @@ TODO -->
         <guilabel>&appname; Preferences</guilabel> screen. You&rsquo;ll want to set both options
         when you start using &app; because if (for example) your accounts are all in Canadian
         Dollars but the generated reports are all in US Dollars, the reports will just say that
-        there are <quote>no data/transactions (or only zeroes) for the selected time period</quote>.
+        there are <quote><computeroutput>no data/transactions (or only zeroes) for the selected time period</computeroutput></quote>.
       </para>
 
       <para>When you create a new account, you have the option to define the commodity in which that account is
@@ -105,19 +108,19 @@ TODO -->
       <para>Since in this example you mostly work in USD, all of the parent accounts are set to USD. Of course,
         if you mostly work in Euros, you could change the currency of these parent accounts to EUR.
         The totals shown in the account tree window will always be converted to the currency of each
-        particular account. Notice, we also set up 3 Starting Balances equity accounts, used to
+        particular account. Notice, we also set up 3 Opening Balances equity accounts, used to
         initially populate the 3 banks.
       </para>
 
       <note>
-        <para>You could also set up just a single Starting Balance account and use a currency transfer to populate
+        <para>You could also set up just a single Opening Balance account and use a currency transfer to populate
           the <quote>different currency</quote> accounts. However, this is more advanced option,
           which is explained in a later section (<xref linkend="currency_purchase2" />).
         </para>
       </note>
 
-      <para>Below you see the result of this example, in which you start with USD 10,000, EUR 10,000 as well as
-        HKD 10,000 in the three bank accounts. Notice that the total of the parent accounts only
+      <para>Below you see the result of this example, in which you start with 10,000 USD, 10,000 EUR as well as
+        10,000 HKD in the three bank accounts. Notice that the total of the parent accounts only
         shows the value of the currency of sub-accounts with matching currencies. In other words,
         the Total Assets and Total Equity values only reflect USD amounts, because &app; has no way
         of evaluating the value of EUR or HKD yet. Once you set up exchange rates between the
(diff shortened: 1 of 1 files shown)


guide/C/ch_currency.xml
```diff
@@ -629,13 +623,14 @@ TODO -->
         <para>You decide to purchase stock in the Beijing Airport (Hong Kong). The ticker for this stock is
           0694.HK on Yahoo! Since you wanted to track all various income and expense amounts, here
           is the necessary account structure:
+        <itemizedlist>
+            <listitem><simpara>Assets:Investments:Brokerage Accounts:Boom:0694.HK (0694.HK)</simpara></listitem>
+            <listitem><simpara>Assets:Investments:Brokerage Accounts:Boom:Bank (HKD)</simpara></listitem>
```
usamimikamen-sk, line 628: I think "Deposit" or "Cash" is better than "Bank".
But it is difficult to change its screenshot right now.

guide/C/ch_currency.xml
```diff
@@ -1353,15 +1342,15 @@ Trading
 
         <para>At this point there have been no trading gains or losses. The balances of the USD and HKD trading
           accounts are the same when converted to USD. The HKD trading account has a debit (-ve
-          balance) and the USD trading account has a credit balance (+ve balance). The nett balance
+          balance) and the USD trading account has a credit balance (+ve balance). The net balance
```
usamimikamen-sk, line 1345: What ve means?

vacuum extraction, vacuum extractor, value engineering, ventilating equipment, vernal equinox, 
vernier engine, visual efficiency, vitamin E, Venezuela, final military Victory in Europe

jralls: I think it's supposed to be positive and negative. @DaveC49, is that correct?

guide/C/ch_currency.xml
```diff
@@ -1342,7 +1331,7 @@ Trading
         </screenshot>
 
         <para>The balances of the Asset and Equity accounts are presented with values set by the exchange rate
-          entered in the transaction to transfer funds on 01/02/2020.
+          entered in the transaction to transfer funds on 02/01/2020.
```
usamimikamen-sk, line 1334: I think it is better to use Feb. 1, 2020.
TODO

jralls: Or 1 Feb 2020. Don't change it to 02/01/2020 unless you make new screenshots to match, it's even more confusing. 

There are quite a few dates on this page that need to be changed and they should all be done in one commit.

usamimikamen-sk: Done

guide/C/ch_currency.xml
fellen: Under Cash I understand coins and bank notes.
How about:
`<!-- ToDo Change on update of screenshot Bank -> Deposit (Type Bank) -->`

fellen (commented):
`<sect2 id="currency_acct1">`

1. The default for currency depends on the region from your environment's LANG (Locale setting). This can be overridden by the pref.
2.  New: There is also in "New Account Hierarchy Setup" a currency selection  defaulting to the Preference, which is then used inside the assistent.
3. @jralls Didn't we want to expose the currency of the root account as the book's currency?

jralls:
> @jralls Didn't we want to expose the currency of the root account as the book's currency somewhere in File->property?

Not quite. We need to make root/book currency selection more certain. It's possible now to create a book without one by creating the CoA by hand, i.e. without the Hierarchy Assistant, and it's also possible to change it with the Hierarchy Assistant on an existing book. Neither are desirable and I don't think we should document that. We also haven't decided how to fix those problems and until we do fix them we can't really document that either.

guide/C/ch_currency.xml
jralls: @fellen, more English ambiguity. "Cash" is commonly used in this context to mean currency assets in a brokerage account. It used to be common in the US for these funds to be invested in a special kind of mutual fund that invested in short-term bonds and maintained a $1 NAV, but since the 2008-9 credit crisis when several of those mutual funds were either unable to maintain the NAV or failed altogether the brokerages have been pushing customers cash account into regular bank accounts. Bank isn't really a bad choice at least for US English.

guide/C/ch_currency.xml
jralls: No response from @DaveC49 so please change +ve to positive and -ve to negative.

jralls:
>`<sect2 id="currency_acct1">`
[edit] but perhaps it should be moved up as a separate section Currency Derivation or similar.
[/edit]

Currency Derivation is too close to Currency Derivatives, of which forex futures might be an example. ISTM the current title for this section, "Account Setup", is accurate: It explains where to change the currency that's set by default when you create a new account.

guide/C/ch_currency.xml
usamimikamen-sk: 79988cac

guide/C/ch_currency.xml
```diff
@@ -28,9 +28,12 @@
       example, you can have a bank account set up in Euros, and another using Hong Kong Dollars.
     </para>
 
-    <para>Some of the issues which arise when using multiple currencies are, how do you transfer funds between
-      accounts with different currencies? How do you calculate the overall value when you have mixed
-      currency accounts? How do reports deal with mixed currencies?
+    <para>Some of the issues which arise when using multiple currencies are as follows:
+    <itemizedlist>
+        <listitem><simpara>how do you transfer funds between accounts with different currencies?</simpara></listitem>
```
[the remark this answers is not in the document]
usamimikamen-sk: b6fd87c7

guide/C/ch_currency.xml
```diff
@@ -105,19 +108,19 @@ TODO -->
...
         if you mostly work in Euros, you could change the currency of these parent accounts to EUR.
         The totals shown in the account tree window will always be converted to the currency of each
-        particular account. Notice, we also set up 3 Starting Balances equity accounts, used to
+        particular account. Notice, we also set up 3 Opening Balances equity accounts, used to
         initially populate the 3 banks.
       </para>
 
       <note>
-        <para>You could also set up just a single Starting Balance account and use a currency transfer to populate
+        <para>You could also set up just a single Opening Balance account and use a currency transfer to populate
```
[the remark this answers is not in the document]
jralls: It's the case generally.  You need the basis cost of any non-home-currency assets in order to be able to have a balanced book with correctly recorded gains when you dispose of those assets. The correct way is to compute the basis of the asset and create a home-currency opening balance for it that you immediately transfer at the basis price to the actual asset account. For fixed assets that should be followed with an accrued depreciation transaction so that the book reflects the current tax value of the asset.

guide/C/ch_currency.xml
usamimikamen-sk: Added comment for translators.

guide/C/ch_currency.xml
```diff
@@ -437,56 +443,44 @@ TODO -->
...
-            <para>Expand the CURRENCY row.
-            </para>
-          </listitem>
+          <step><simpara>Expand the <guilabel>CURRENCY</guilabel> row.</simpara></step>
 
-          <listitem>
-            <para>Double click on the currency for which you want to disable exchange rate retrieval.
-            </para>
-          </listitem>
+          <step><simpara>Double click on the currency for which you want to disable exchange rate retrieval.</simpara></step>
```
fellen, line 455: I was long searching. Either we should 

1. display the column in the program by default or
2. Remind the reader to display the "Get Quotes" column.

@jralls et al.: your opinion?

fellen (commented):
As I currently have little time the interim state of a "play through".
I made updated screenshots, but will wait with committing for the "Get Quotes" column question.

fellen:
Merging the current state.
Keep in mind the future changes.

Thanks

PR #202, opened 2021-07-23
Proposal: Move chapter Depreciation before Other Assets.

usamimikamen-sk:
I read both the Depreciation chapter and the Other Assets chapter.
The Depreciation chapter focus on the topic of depreciation only, and the usages of GnuCash on depreciation is described clearly. 
On the other hand, the Other Assets chapter describes many assets and they are a little abstract.

So I think it is better that we move the Depreciation capter before the Other Assets chapter.

### DIFF
--- a/guide/C/ch_oview.xml
+++ b/guide/C/ch_oview.xml
@@ -449,12 +449,12 @@
         </listitem>
 
         <listitem>
-          <para><xref linkend="chapter_other_assets"></xref>
+          <para><xref linkend="chapter_dep"></xref>
           </para>
         </listitem>
 
         <listitem>
-          <para><xref linkend="chapter_dep"></xref>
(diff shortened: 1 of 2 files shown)


fellen:
It is related to Investments:Selling Shares and closer to Capital Gains. There we have already revaluations.

From my POV a logical sequence would be
Currencies,
Investments,
Capital Gains & Depreciations
(all about entering Data and finally)
Reports

There has been a long discussion on https://bugs.gnucash.org/show_bug.cgi?id=687820 and related bug reports. But it missed the original division between 
A: Tech. Ref.: Help,
B.1: didactical Tutorial
B.2: task oriented Guide

usamimikamen-sk:
I see, I'm going to close this Pull Request.

fellen:
Any plans of an improved  redo?

PR #203, opened 2021-08-02
Improved PDF creation with the "variablelist" element

CWehli:
Using the parameter "variablelist.as.blocks" the term and its
description are not printed on the same line. The term itself hangs to
the left and the description is printed one line below.


CWehli:
PR #204 contains a more promising and comprehensive approach.

PR #204, opened 2021-08-04
WIP: restructure xsl

usamimikamen-sk:
test for running check process.


fellen:
Can you explain, what you are trying to do?
Until now we had in xsl the customization (mostly an ancient gnome-doc template)
and in 1.79.2 the official (distribution neutral) docbook-xsl allowing an easy update.

Watch also @CWehli is trying to fix FOP's varlist issue in PR #203. Some coordination is required.
BTW FO processing has a low priority (at least for me). Reducing redundancy between both documents seems more important—also requiring your coordination.

CWehli:
I really like the approach of restructuring the XSL folder. 
However, my skills in this area are very limited, so I will close PR #203.

usamimikamen-sk:
I'm working on fixing some problem.

(1) HTML and FO xsl stylesheet fix
There are 4 file formats that are converted by xsltproc and xsl stylesheet.
HTML, CHM and EPUB are based on the HTML. PDF is based on FO.  
And the difference between HTML and FO is absorbed by xsl stylesheet.

So if we want to customize the output, some of elements should be treated
differently. Some block elements like titlepages and variablelst//term XPath
elements are especially.

In gnucash-docs, general-customization.xsl and other xsl files in the xsl drectory play the role.
But they are insufficient and cause some errors.

Therefore, I'm cleaning them up.

New main structure is as follows:
```
gnc-custom-html.xsl --- gnc-custom-common.xsl ---------- OASIS html xsl
                     |          ^                  |
gnc-custom-pdf.xsl --           |                   ---- OASIS FO xsl
                                |            
       ^                       common  
       |                        guimenu.xsl
   format specific              emphasis.xsl
     gnc-titlepage-fo.xsl
     gnc-titlepage-html.xsl

```

(2) Remove unused customization and workaround files
Some customization files are not used now. Some workaround are already merged into the OASIS xsl.
So I remove such files.


(3) Flatten the depth of xsl directory
When I started the customization, I needed to flatten the directory structure in order
to import files properly. l10n.xml is an example.

But now I found l10n.xml is not necessary. So even if I restore the directory structure, it may work.
As for upgrading OASIS xsl in the new structure, we only need to overwrite them except xsl/gnucash directory.


(4) Modify cmake
I modified cmake files because I changed xsl directory structure.


(5) Bug fix and improvements
 * variablelist/varlistentry/term fix. Now  `<team>a</team><team>b</team>` works properly
 * PDF admon graphics fix
 * Import PR #203 @CWehli into gnc-custom-pdf.xsl
 * EPUB admon grafics fix
 * PDF paper size fix, USletter and A4 is applied to locale C and the others, respectively
 * Improve HTML (EPUB) and PDF titlepage
 * Fix FDL1.1 display problem 
 They are only in guide/C now.


(6) Further plan
I'm planning to change dtd structure in order to translate ENTITY.
For example, `<keycap function="control">Ctrl</keycap>` in Engish can be translated into `<keycap function="control">Strg</keycap>` in German.

usamimikamen-sk:
distribution neutral) docbook-xsl allowing an easy update.
> [Edit] I am not against applying useful patches from distributions, but they should be documented in changelog or at least README. [/Edit]

It is written in xsl/README-GnuCash file.

jralls:
And ChangeLog is generated from `git log`. If the commit message mentions the source of the patch then it will be in the next release's ChangeLog.

The Readme-Gnucash change says
>These XSLs are borrowed from Fedora 34.
If you'd like to customize the styles for GnuCash , edit files in the gnucash/ directory.

That strikes me as not quite correct: Most of the stylesheets are from OASIS. It's unclear in this commit whether you've updated those or just moved them from the versioned directory. Either way that should be a separate commit whose message says what you did. It would be better to name the directory containing those stylesheets `borrowed/oasis-group` instead of just `stylesheets`. To make clear that it's a subset of https://github.com/docbook/xslt10-stylesheets I think preserving `xsl` in the directory structure, so `borrowed/oasis-group/xsl`. Since you've pruned the directory to just the stylesheets we use it's probably not appropriate to include the OASIS Readme: Most of the files it refers to aren't present. However you've removed the OASIS COPYING and AUTHORS files. Their license is not GDPL and must be included with their files.

If there are customized stylesheets from the Fedora project ***And RedHat has licensed them under a GDPL-compatible license*** then those should be added *unchanged* in their own commit to a separate `borrowed/fedora` directory both to ensure that the RedHat copyright is correctly acknowledged and to make it easy to update them. That commit message should explain where they came from and why they're useful. If the license isn't GDPL-compatible then we can't use them. In any case the RedHat LICENSE or COPYING and AUTHORS files should be in the directory.

Then any work that you've done yourself should be in one or more separate commits grouped by purpose with an explanatory commit message for each.

fellen:
> HTML, CHM and EPUB are based on the HTML. PDF is based on FO.

I don't see that EPUB is based on HTML,
We have also MOBI,which is based on EPUB.

jralls:
> I don't see that EPUB is based on HTML,

Right. It isn't based on html, [it is html](https://en.wikipedia.org/wiki/EPUB#Contents).

fellen:
> Right. It isn't based on html, [it is html](https://en.wikipedia.org/wiki/EPUB#Contents).

No it is XHTML + container, …
There are slightly differences like writing of attributes,…

jralls:
Read the link. It's HTML5 now.

fellen:
> Read the link. It's HTML5 now.

But nobody had updated the abstract there. ;-)

usamimikamen-sk:
> The Readme-Gnucash change says
> 
> > These XSLs are borrowed from Fedora 34.
> > If you'd like to customize the styles for GnuCash , edit files in the gnucash/ directory.
> 
> That strikes me as not quite correct: Most of the stylesheets are from OASIS. It's unclear in this commit whether you've updated those or just moved them from the versioned directory. Either way that should be a separate commit whose message says what you did. It would be better to name the directory containing those stylesheets `borrowed/oasis-group` instead of just `stylesheets`. To make clear that it's a subset of https://github.com/docbook/xslt10-stylesheets I think preserving `xsl` in the directory structure, so `borrowed/oasis-group/xsl`. Since you've pruned the directory to just the stylesheets we use it's probably not appropriate to include the OASIS Readme: Most of the files it refers to aren't present. However you've removed the OASIS COPYING and AUTHORS files. Their license is not GDPL and must be included with their files.

I see.
I imported from xsl stylesheet files from Fedora because Changelog saids '2018-10-12 Geert Janssens * Add DocBook xsl v1.79.2 (taken from Fedora 27 installation)'.

If there are license problems exists, it is better to replace them with pure OASIS stylesheets sooner.

jralls:
There's no license issue *per se*, it's just that the OASIS license isn't GDPL so their copyright notice and license need to be in the directory with the stylesheets regardless of how they get added.

As a separate matter I'm not all convinced that we need the stylesheets in our repository at all. If we declare them a dependency instead then we don't have to update them, we need only worry about keeping our customizations up to date.

usamimikamen-sk:
> > Read the link. It's HTML5 now.
> 
> But nobody had updated the abstract there. ;-)

As for GnuCash docs EPUB, current version is 2. And output files are based on xhtml.

EPUB 3 is mainly based on html5 in order to support right-to-left scripts such as Arabic and many non-latin typographic features, but can be used xhtml for compatibility.

OASIS xslt10-stylesheets uses xhtml 1.1 for EPUB 2 and html5 for EPUB 3.

usamimikamen-sk:
> There's no license issue _per se_, it's just that the OASIS license isn't GDPL so their copyright notice and license need to be in the directory with the stylesheets regardless of how they get added.
> 
> As a separate matter I'm not all convinced that we need the stylesheets in our repository at all. If we declare them a dependency instead then we don't have to update them, we need only worry about keeping our customizations up to date.

@jralls seems to have the same opinion with @fellen. If so, we have to change CMakeLists.txt in order to search packages.
I will try it but it seems to be difficult to absorb the difference between Windows, Mac and Linux distributions.

usamimikamen-sk:
> @jralls seems to have the same opinion with @fellen. If so, we have to change CMakeLists.txt in order to search packages.
> I will try it but it seems to be difficult to absorb the difference between Windows, Mac and Linux distributions.

I found the way to use system xsl stylesheets and dtds, but it is necessary to change across gnucash-on-windows, gnucash-on-osx and gnucash-docs repositories.
Therefore I decide that I'll import AUTHORS, LICENCE and Changelog files from RedHat/OASIS into this PR.
I will create other PRs when I'll ready to use system DTD and xsl stylesheets.

usamimikamen-sk:
> As a separate matter I'm not all convinced that we need the stylesheets in our repository at all. If we declare them a dependency instead then we don't have to update them, we need only worry about keeping our customizations up to date.

I wonder whether docbook-dtd and docbook-xsl packages are installed in the 6 check processing environments in GitHub.  
Are they installed in these environments? 
Can I add xmllint options, such as --nonet and --catalog?

jralls:
I don't think that they are but for the Ubuntu ones it's simple to add via the package manager.

usamimikamen-sk:
On Windows:
$ pacman -F /etc/xml/
etc/xml/ is owned by msys/docbook-xml 4.5-2
etc/xml/ is owned by msys/docbook-xsl 1.79.2-1

cmake/AddPdfTarget.cmake
```diff
@@ -6,32 +6,21 @@ function (add_pdf_target docname lang entities figures)
...
-    # * all others use A4.
-    set (XSLTFLAGS_FO "--stringparam paper.type A4")
-    if (lang STREQUAL "C")
-        # For the fallback language determine paper format depending on locale
-        # Only US or C will be set to letter. All others use A4.
-        set (ENV_LANG $ENV{LANG})
-        if (ENV_LANG AND ENV_LANG MATCHES ".*_us.*|C") # Replacing ENV_LANG here with if ($ENV{LANG}) won't work.
-            set (XSLTFLAGS_FO "--stringparam paper.type letter")
-        endif()
-    endif()
```
gjanssens, line 20: While I like you moved this inside the stylesheet it now no longer does exactly the same as was done here.

The assumption behind the logic above is that only native English speaking countries use Legal as paper format. All others default to A4 (this may be a false assumption by the way, just explaining the rationale that was used here).
So for all languages other than English ("C"), the choice is straightforward: use A4.
"C"/English is special because we consider that the fallback language. That is, if someone natively speaking a language for which we don't have the documentation, that person is assumed to go for English. Add to that the English edition is the most complete one so the most preferred version in any case. As said before most of the world uses A4 formatting, so unless the user choosing English because said user is native English, the expectation is to get a A4 formatted document. This final subtlety is lost in your change.

Now having explained all that, I think the above logic is probably trying to be too smart also. It would be much better to take what you have written in the stylesheet as default but allow the user a way to override by providing a PDF_PAPER_FORMAT cmake option. With that a user could run
cmake -DPDF_PAPER_FORMAT=A4 <src-dir>
or
cmake -DPDF_PAPER_FORMAT=Letter
in cases our simplified algo doesn't do what the user wants.

guide/C/.gitignore
```diff
@@ -1,3 +1,4 @@
 *.pdf
 gnucash-guide
 gnucash-help
+stylesheet/
```
gjanssens, line 4: This is inconsistent. Either we do this in all .gitignore files or we do it nowhere.

To be honest I have rather become a proponent of teaching users to build outside of the source directory if the want to do development work. That is I don't care if they're just  building for their own personal use or for packaging, but if a user intends to contribute to the source tree (and hence needs to make git commits or patches), the build should happen outside of that source tree to avoid commit pollution.

In that respect most of the .gitignore entries should rather be removed than adding more to it.

gjanssens: Also the commit message is not really covering what's happening here. Is this a fixup you intended to squash later ?

gjanssens (commented):
This is quite a good and much needed cleanup! I have a few remarks and points for discussion added inbetween.

guide/C/.gitignore
usamimikamen-sk: > This is inconsistent. Either we do this in all .gitignore files or we do it nowhere.
> 
> To be honest I have rather become a proponent of teaching users to build outside of the source directory if the want to do development work. That is I don't care if they're just building for their own personal use or for packaging, but if a user intends to contribute to the source tree (and hence needs to make git commits or patches), the build should happen outside of that source tree to avoid commit pollution.
> 
> In that respect most of the .gitignore entries should rather be removed than adding more to it.

This is a workaround for the behavior of fop.
The stylesheet/ must be in the same directory of the main XML file like gnucash-guide.xml in order to include admonition graphics properly.
Therefore I added
```
cmake/AddPdfTarget.cmake line 23:
COMMAND ${CMAKE_COMMAND} -E copy_directory "${CMAKE_SOURCE_DIR}/stylesheet" "${CMAKE_CURRENT_SOURCE_DIR}/stylesheet"
```
and added it in .gitginore.

Another way is that I copy all the XML and image files into builddir and run xsltproc there.
Do you think which is better?

cmake/MakeChm.cmake
```diff
@@ -66,4 +66,4 @@ execute_process(
     ERROR_QUIET
 )
 
-file(COPY "${htmlhelpdir}/${docname}.chm" DESTINATION "${BUILD_DIR}")
+execute_process(COMMAND ${CMAKE_COMMAND} -E copy "${htmlhelpdir}/${docname}.chm" "${BUILD_DIR}")
```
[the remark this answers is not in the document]
usamimikamen-sk: There are some reasons.

* I'm not familiar with cmake and ninja. 
* In other locations, execute_process is used for file copy and directory copy. Therefore I'd like to unify them.
* I prefer "cp src dest" format. And I don't like "cp src DESTINATION dest" format.

Would you tell me good manners and commands to copy files and directories in the cmake?
I'd like to follow the best practice of cmake.

cmake/AddPdfTarget.cmake
usamimikamen-sk: I see, but I'm reading cmake and ninja documents now. So it is difficult for me to fix it right now.

guide/C/.gitignore
usamimikamen-sk: I found another option.
If you don't hesitate to use symbolic links, they are enough because PDF/FOP seems to be supported only on the UNIX.

usamimikamen-sk:
@jralls @gjanssens 

I looked up the following 2 sites:
https://github.com/Gnucash/gnucash-on-osx
https://gitlab.gnome.org/GNOME/gtk-osx/

The docbook-dtds and docbook-xsls are not included either.
I don't have Mac OS environment, so I can't create patches for jhbuild.
Would you try it on Mac OS?

guide/C/.gitignore
gjanssens: From what I  can see the path to the admonitions images is parameterized (via `$admon.graphics.path`). Have you tried defining that parameter on the xlstproc command line and have it point at `${CMAKE_SOURCE_DIR}/stylesheet` rather than hard-coding it in gnc-custom-common.xsl ?

As far as I'm concerned that directory can even be renamed (like admonitions or something like that) and  be put in the xsl  directory.

cmake/AddPdfTarget.cmake
gjanssens: It's a minor issue IMO. Can be revised later.

cmake/MakeChm.cmake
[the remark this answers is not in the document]
gjanssens: I think `file(COPY ....)` is preferred over `execute_process` at least in this  case.

`file(COPY ...)` is a built-in command, executed in-process by cmake.  IIRC it also tracks file changes and will automatically update copied files in case any of the source files changes.

`execute_process` on the other hand will tell cmake to run an external command. In this  case that command happens to be cmake again, but this  case in script mode to call the builtin script command `copy`. Aside from the unwanted recursion, the main disadvantage of this mode is that the calling cmake command doesn't know anything of the changes made by the called command, so we lose tracking of source file changes. The script mode of cmake is to be used when cmake commands should be called at build time (that is while make/ninja is running) rather than during cmake's configuration or generation phases or as part of custom commands if not other option is available.

In retrospect this probably goes for all the places where you changed `file(COPY ...)` to a `COMMAND cmake -E copy` statement.

guide/C/.gitignore
usamimikamen-sk: > From what I can see the path to the admonitions images is parameterized (via `$admon.graphics.path`). Have you tried defining that parameter on the xlstproc command line and have it point at `${CMAKE_SOURCE_DIR}/stylesheet` rather than hard-coding it in gnc-custom-common.xsl ?

I haven't come up with it until now. I just followed the previous method like html. I'll try it.

> As far as I'm concerned that directory can even be renamed (like admonitions or something like that) and be put in the xsl directory.

If you developers agree with moving stylesheet/ directory, I want to change its name too.

cmake/MakeChm.cmake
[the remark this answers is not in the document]
usamimikamen-sk: OK. I will look up the cmake manual and unify the procedures.

@gjanssens 
On another topic, why are the files split into MakeChm.cmake and AddChmTraget.cmake for chm?
I'd like to merge them into AddChmTraget.cmake. Do you have some technical problems?

jralls:
A subset of the docbook stylesheets [are included](https://gitlab.gnome.org/GNOME/gtk-osx/-/blob/master/modulesets-stable/bootstrap.modules#L123) but they're obsolete. That was originally intended to provide only what was needed for gtk-doc 12 years ago. I'll bring it up to date.

jralls:
I updated the gtk-osx-docbook item in bootstrap.modules and now this branch builds correctly on macOS.

cmake/AddPdfTarget.cmake
fellen: > While I like you moved this inside the stylesheet it now no longer does exactly the same as was done here.

Perhaps we should ask the user's OS on initialization and then use a string-parameter to overwrite A4?.

> The assumption behind the logic above is that only native English speaking countries use Legal as paper format. All others default to A4 (this may be a false assumption by the way, just explaining the rationale that was used here).
> So for all languages other than English ("C"), the choice is straightforward: use A4.

Not exactly: in en_GB it should be A4. CA's P series is a rounding of US formats to 5 mm. And some es speaking countries in the US "backyard" use letter called „Carta“ there.

https://en.wikipedia.org/wiki/Paper_size
https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/territory_information.html

guide/C/.gitignore
fellen: IMHO we should drop the stylesheet directory as far as possible. The admonition images are also in xslt/images like the missing admonitons.
I suspect it was not the case when the project started.

xsl/README
```diff
@@ -1,5 +1,9 @@
 The gnucash-docs now use system xslt stylesheets and there are some customization files
-in this directory. But only images/* contents are left to make admonition and callout graphics.
+in this directory. But only images/ contents are left to make admonition and callout graphics.
+
+Additionally, the following image files of GnuCash are put into images/
+directory in convenience:
+ - gnucash-icon.png
```
fellen, line 6: Tihis is a very old gnucash-icon, mentioned in NEWS: 1.7.7  - 05 January 2003

I see hardcoded ocurrences here:
```
./xsl/titlepage.xml:29:    <img align ="left" src="stylesheet/gnucash-icon.png" alt="GnuCash Logo"/>
./xsl/titlepage.xml:57:    <img align ="left" src="stylesheet/gnucash-icon.png" alt="GnuCash Logo"/>
./xsl/titlepage.xsl:82:    <img align="left" src="stylesheet/gnucash-icon.png" alt="GnuCash Logo"/>
./xsl/titlepage.xsl:259:    <img align="left" src="stylesheet/gnucash-icon.png" alt="GnuCash Logo"/>
```

We should replace it by one from https://github.com/gjanssens/gnucash/tree/maint/data/pixmaps.
IMHO for the titlepage gnucash_splash.png would be best.

For metadata like .destop, .rpm … we should create one based that in highcolor—perhaps overlayed by the yellow Help question mark as @gjanssens did for other components of he project.

xsl/images/colorsvg/caution.svg
```diff
@@ -0,0 +1,141 @@
+<?xml version="1.0" encoding="utf-8"?>
```
fellen, line 1: Your whole xsl/images changes:
No, we do not want to replace the current OASIS images by our 20 years old versions!

fellen: The check errors happen because there are still references to stylesheet.
Just click details in the failed checks.

usamimikamen-sk: > Your whole xsl/images changes:
> No, we do not want to replace the current OASIS images by our 20 years old versions!

The resolution of current icons in stylesheets/ is too low for PDF files.
Do you have good SVG files or high-resolution files?

usamimikamen-sk: https://github.com/gjanssens/gnucash/blob/maint/data/pixmaps/hicolor/scalable/apps/gnucash-icon.svg
is good.

xsl/README
usamimikamen-sk: > We should replace it by one from https://github.com/gjanssens/gnucash/tree/maint/data/pixmaps.
> IMHO for the titlepage gnucash_splash.png would be best.

It is very difficult for me to change the layout of the titlepage, especially FO/PDF.

xsl/images/colorsvg/caution.svg
usamimikamen-sk: > The check errors happen because there are still references to stylesheet.
> Just click details in the failed checks.

I see.
I'm trying to change CMakeList.txt and CMakefiles/ right now.

xsl/README
fellen: Oops, I had not seen that I still was on a clone. I meant:
https://github.com/Gnucash/gnucash/tree/maint/data/pixmaps/

xsl/images/colorsvg/caution.svg
usamimikamen-sk: > The resolution of current icons in stylesheets/ is too low for PDF files.
> Do you have good SVG files or high-resolution files?

I'll change to use png for HTML, EPUB, CHM because 24x24 png is enough to display.
But I'd like to use SVG for PDF because 96x96 png is necessary.

guide/C/.gitignore
usamimikamen-sk: > In that respect most of the .gitignore entries should rather be removed than adding more to it.

I removed {guide,help}/${lang}/.gitignore completely.

cmake/MakeChm.cmake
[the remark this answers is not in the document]
usamimikamen-sk: I cleaned up AddChmTraget.cmake and removed MakeChm.cmake.

cmake/AddPdfTarget.cmake
usamimikamen-sk: ToDo.

xsl/README
usamimikamen-sk: Done.

> IMHO for the titlepage gnucash_splash.png would be best.

wontfix now

usamimikamen-sk:
> On Windows:
> $ pacman -F /etc/xml/
> etc/xml/ is owned by msys/docbook-xml 4.5-2
> etc/xml/ is owned by msys/docbook-xsl 1.79.2-1

TODO:
change Gnucash/gnucash-on-windows to install mingw32 docbook-xsl

usamimikamen-sk:
I squashed the commits.
```
(1) restructure docbook xsl stylesheets.
  * The xsl/1.79.2/ was completely removed and gnucash-docs uses the system docbook-xsl stylesheets now.
  * The images of OASIS docbook-xsl was copied into xsl/images.
  * The admonition icons under stylesheets/ were merged into xsl/images/.
  * The stylesheets/ entry for distfiles in Makefile.am was removed.
  * The old GnuCash icons were updated and put into xsl/images/.
  * The XSL stylesheets for GnuCash customization were split.
       First, gnc-custom-chm.xsl, gnc-custom-epub.xsl, gnc-custom-fo.xsl
           and gnc-custom-html.xsl are for CHM, EPUB, FO/PDF and html, respectively.
       Then, gnc-custom-common.xsl is imported for common use.
  * Unused xls stylesheets were removed.
  * The variablelist.xsl was split into variablelist-fo.xsl and
    variablelist-html.xsl because they depend on formats.
  * The stylesheets for title page were cusomized.
  * xsl/README was updated.
  * For PDF:
     - PR #203 by @CWehli: The variable list customization was imported.
     - The paper size is set to USletter and A4 for C and the other locales respectively.
     - Other minor improvements were applied.
  * For CHM:
     - The Character code is set Shift_JIS and KOI8-R for ja and ru locale respectively.

(2) cmake related
  * The stylesheet/ was removed from distfiles in CMakeList.txt.
  * Usages comment of function was added at the top of the files.
  * AddGncDocTargets.cmake:
     - SVG files were removed from the list "${figure}" because they aren't used right now.
     - "${dtd_files}" was added to support multiple dtd files.
  * AddChmTarget.cmake:
     - MakeChm.cmake was merged.
     - The working directory was changed to "${CMAKE_CURRENT_BINARY_DIR}/htmlhelp".
     - The methods of copying figures and icons were changed to use add_custom_command.
     - Dependencies on "${dtd_files}" were added.
     - The dependencies were rearranged.
     - Unused functionalities in GnuCash, such as alias and map, were removed.
     - The "--stringparam gnc.lang ${lang}" argument is added to the xsltproc so as to swich the character codes.
     - The dtd file name for xsltproc was changed to support the new xsl/ structure.
   * AddEpubTarget.cmake:
     - The dtd file name for xsltproc was changed to support the new xsl/ structure.
     - Dependencies on "${dtd_files}" were added.
     - The admonitions and Gnucash icons were added.
   * AddGHelpTarget.cmake:
     - The dtd file name for xsltproc was changed to support the new xsl/ structure.
     - Dependencies on "${dtd_files}" were added.
   * AddHtmlTarget.cmake:
     - The dtd file name for xsltproc was changed to support the new xsl/ structure.
     - Dependencies on "${dtd_files}" were added.
     - The admonitions and Gnucash icons were added.
   * AddPdfTarget.cmake:
     - The working directory was changed to "${CMAKE_CURRENT_BINARY_DIR}".
     - fop.xconf.in and guide/ja/CMakeList.txt were changed to support the new working directory.
     - The dtd file name for xsltproc was changed to support the new xsl/ structure.
     - Dependencies on "${dtd_files}" were added.
     - The admonitions and Gnucash icons were added.
   * MakeChm.cmake was removed.

(3) .gitignore related
   * {guide,help}/*/.gitignore were deleted completely.

(4) FDL 1.1  and legal related
   * The markups of fdl-appendix.xml were fixed for all languages.
   * legal.xml and legal-disclaimer.xml were unified for all languages.
   * legal info (legal.xml) and disclaimer (legal-disclaimer.xml) were moved into the preface.
   * distfiles entries in Makefile.am and CMakeList.txt were modified.

(5) others
* README: Garbage characters were removed.
* gnucash-docs.spec.in: The License line was changed to FLD 1.1 or later.
* AUTHORS: Move me to active users.
```

usamimikamen-sk:
Sigh, some PNG icons are corrupted.
TODO.

fellen:
@jralls In theory we say dependency changes only major releases.
But looking at https://repology.org/project/docbook-xsl/versions all important distributions seem to have docbook-xsl. Some BSDs seem slighty behind.
Because https://cdn.docbook.org/release/xsl-nons/1.79.2/release-notes#V1.79.2 says `This release doesn't contain any functional changes to the stylesheets.` I would suggest to require docbook-xsl>=1.78.1—at least for publishers.
Then an exception would be possible. Your opinion?

fellen:
@usamimikamen-sk 666 changed files is a complex beast. Can you split it in the pure "Require docbook-xsl" with

1. dropping our copy, 
2. adjust build systems *make, configure files
3. and metatdata like README, rpm, …

and a separate PR with your improvements?

xsl/gnc-custom-chm.xsl
```diff
@@ -0,0 +1,66 @@
...
+<!-- language specific character code for wchar
+     ja: Shift_JIS
+     ru: KOI8-R
+-->
+
+<xsl:param name="htmlhelp.encoding">
+  <xsl:choose>
+    <xsl:when test="$gnc.lang = 'ja'">Shift_JIS</xsl:when>
+    <xsl:when test="$gnc.lang = 'ru'">KOI8-R</xsl:when>
+    <xsl:otherwise>ISO-8859-1</xsl:otherwise>
```
fellen, line 33: ISO-8859-1 is missing € and it is often used.
Can htmlhelp not use UTF8?

usamimikamen-sk: Yes and no.

The UTF-8 may work in the HTML pane because "iexplore.exe" renders it. But HTML HELP renders the TOC pane.
THE HTML HELP uses legacy MS codepages, such as 932(ja), 855(ru), 1252(latin). So the EURO sign may not be rendered properly.

The current default setting for CHM is ISO-8859-1 and I can't expect what happens if I set it UTF-8.

usamimikamen-sk:
> @usamimikamen-sk 666 changed files is a complex beast. Can you split it in the pure "Require docbook-xsl" with
> 1. dropping our copy,
> 2. adjust build systems *make, configure files
> 3. and metatdata like README, rpm, …




(1) 31ac654ca7bedd3f1
 581 files changed, 4370 insertions(+), 290595 deletions(-)
(2)  e64318e756934003c7b47e6042013fabac3a1229
10 files changed, 195 insertions(+), 165 deletions(-)

I can't split 1 and 2 because changing the paths affects the build system.   

See the output of "git log --stat 31ac654ca7bedd3f1e48dd8 -1".

fellen:
> I can't split 1 and 2 because changing the paths affects the build system.

My numbered list was not meant as commits. It served only as description of the kinds of changes, which belong to it.
In contrary: Each commit should build errorfree.

xsl/gnc-custom-chm.xsl
jralls: > THE HTML HELP uses legacy MS codepages, such as 932(ja), 855(ru), 1252(latin). So the EURO sign may not be rendered properly.

It's about 12 years past time to get rid of the chm build and open the documentation in the default browser or in a gtkwebkitwebview. That's obviously out of scope for this PR and not something that I'd ask of you.

usamimikamen-sk:
> > I can't split 1 and 2 because changing the paths affects the build system.
> 
> My numbered list was not meant as commits. It served only as description of the kinds of changes, which belong to it.
> In contrary: Each commit should build errorfree.

I didn't have enough words.

(1) (2)
PR #211 is atomic. 

Second, create:
(4) FDL 1.1 and legal related

Then create:
(3) .gitignore related
(5) others

usamimikamen-sk:
> > Current Status of cmake:
> > ```
> > * `cmake --build .`  -> Do Nothing
> > ```
> 
> That was deliberately so. On linux the help and guide are parsed by yelp which can simply take our xml sources as input. We don't have to do any xsl transformations on those. Hence on linux there's nothing to build.

I see.


> > ```
> > * `cmake --install .` -> Don't work.
> > ```
> 
> What errors do you get ?

I tried again, and errors are not reproduced.
Thank you for your comment.

cmake/AddChmTarget.cmake
```diff
@@ -1,36 +1,75 @@
...
+    # Copy figures for this document
+    file(MAKE_DIRECTORY "${BUILD_DIR}/figures")
+    add_custom_command(
+        OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/chmfigtrigger"
+        COMMAND ${CMAKE_COMMAND} -E copy ${figures} "${BUILD_DIR}/figures"
+        COMMAND touch "${CMAKE_CURRENT_BINARY_DIR}/figtrigger"
+        DEPENDS ${figures})
+
+    # Copy icon files for this documents
+    # Removed dependency because icon files are rarely changed.
```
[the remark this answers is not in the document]
usamimikamen-sk: > > I would not omit the dependency just because this rarely changes. It is still dependent on it and it may change. Not setting the dependency would in that case force users to reset their build system.
> 
> @gjanssens
> I rethought and I'll do it in another PR.

I made PR #215 to add a dependency to GnuCash icon files.

cmake/MakeChm.cmake
```diff
@@ -1,69 +0,0 @@
...
-            list(APPEND HHMAP "${sectid}=${count}")
-        endif()
-    endforeach()
-endforeach()
-
-set(HHP "${HHP};${MAP}")
-string(REPLACE ";" "\n" HHP_OUT "${HHP}")
-string(REPLACE ";" "\n" HHMAP_OUT "${HHMAP}")
-file(APPEND "${htmlhelpdir}/htmlhelp.hhp" ${HHP_OUT})
-file(WRITE "${BUILD_DIR}/${mapfile}" ${HHMAP_OUT})
```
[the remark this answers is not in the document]
gjanssens: I am aware  this PR is closed and you will resubmit this work in a future PR. But just for the record I am not convinced thhe code to generate the hhp and hhmap files  can just be  discarded. GnuCash parses the hhmap file in gnucash/gnome-utils/gnc-help-utils.c. So this change should be tested a bit more before it can be applied. Be sure not to mix this in the same commit t hat will drop the xsl stylesheet from gnucash-docs so they can be reviewed independently.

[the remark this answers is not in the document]
gjanssens: > GnuCash doesn't `#include <htmlhelp.h>` when it is build.

It does include this, in the same gnc-help-utils.c file.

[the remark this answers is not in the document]
usamimikamen-sk: It is my oversight.

Then, where is "context.h" and "alias.h"? I can't find them. They are used for context-sensitive help. For example, when we press the "HELP" button on the Scheduled Transaction window, if it exists, the Scheduled Transaction page of GnuCash-help will open. And it is not necessary to create [map] and [alias] sections by our hands.

Just copy them as "context.h" and "alias.h", and run `make chm`, [map] and [alias] section will be created automatically by xsltproc.

PR #208, opened 2021-08-20
remove xslt stylesheet from xsl directory

usamimikamen-sk:
I removed xslt stylesheets from xsl directory.
I left GnuCash specific customization files and images/*.


usamimikamen-sk:
I'll merge this PR into #204.

PR #210, opened 2021-08-22
Remove obsolete appendices (VAT, FAQ) from GnuCash Guide.

usamimikamen-sk:
I removed obsolete appendices (VAT, FAQ) from GnuCash Guide.
The files are appendixb.xml and appendixc.xml.

I'd like to move AUX format (appendix d)  to GnuCash Help in another Pull Request.


fellen:
Thanks!

About the future move of appendix d:

I see currently the problem of moving translations without coverage:
**ru** file is untranslated. can be dropped
**ja**: perhaps you want to create a template before?
**it**: no separate file,
-  somehow we should be able to move it as section from guide to help—like in my last commit
- patching the po file would be an option for a new po-based translator, but not for us.

**C, de, pt** remain for normal move and adjust any make files
about new filename: I suggest app_chk

de: there is another appendixe: command line options and menu shortcuts 
@Mechtilde do you remember the origin?
@CWehli do you have some use for it?

Another issue here: we should drop the stupid Help_ prefix from the filenames in the help directory.

CWehli:
> de: there is another appendixe: command line options and menu shortcuts
> @CWehli do you have some use for it?

@fellen: I'll make myself a copy and put this on the big wish list for the further improvement of the help documentation.

PR #213, opened 2021-08-25
rm  {guide,help}/*/.gitignore

usamimikamen-sk:
* {guide,help}/*/.gitignore were deleted completely.

Split from PR #204.

### DIFF
--- a/guide/C/.gitignore
+++ b/guide/C/.gitignore
@@ -1,3 +1,0 @@
-*.pdf
-gnucash-guide
-gnucash-help
--- a/guide/de/.gitignore
+++ b/guide/de/.gitignore
@@ -1,3 +1,0 @@
-*.pdf
-gnucash-guide
-gnucash-help
--- a/guide/it/.gitignore
+++ b/guide/it/.gitignore
@@ -1,3 +1,0 @@
-*.pdf
-gnucash-guide
-gnucash-help
--- a/guide/ja/.gitignore
+++ b/guide/ja/.gitignore
@@ -1,3 +1,0 @@
-*.pdf
-gnucash-guide
-gnucash-help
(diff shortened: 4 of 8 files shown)


fellen:
> @gjanssens told @usamimikamen-sk on a different PR to remove all of the build products because one shouldn't be building in the source directory anyway. We need to keep the autotools source directory litter like Makefile.in until we get rid of autotools and it's helpful to ignore various editor artifacts like emacs's *.~, though one could argue that contributors should have their chosen toolset's litter in their ~/.gitignore so that it's not needed in project ones.

Did you mean we should keep the current dirty file or drop it?

IMHO we should keep a small clean file for most cases instead of sending potential users at first to read https://git-scm.com/docs/gitignore and configure their own setup. Perhaps they will never return because they run out of time.

jralls:
> Did you mean we should keep the current dirty file or drop it?

We need to keep a file that ignores the files that autotools inserts in the source directory even for out-of-source builds until we remove the autotools build; after that we won't need .gitignore in the repository.

> IMHO we should keep a small clean file for most cases instead of sending potential users at first to read https://git-scm.com/docs/gitignore and configure their own setup. Perhaps they will never return because they run out of time.

.gitignore is a convenience to keep the noise down when one runs `git status`; you can run `git status -uno` and filter out all of the files that aren't under version control. If the build tools don't make any noise--and Cmake doesn't as long as you don't build in the source directory--then it's not appropriate to have a standard one. It's also not possible to create a an encyclopedic one that ignores every editor's backup files. That's better left to the users to put their editor's backup file pattern in their ~/.gitignore since most people will use only one or two editors and they'll know the pattern to use. Consider as well that git-aware editors don't make backup files in git repos because there's no need to.

fellen:
Thanks!

@jralls commit 93ec8aa adds only the entries in the source tree by autotools.

PR #216, opened 2021-09-03
(5a)Rename directory name help/  to manual/ in order to fix Bug#797622.

usamimikamen-sk:
Depends on #217
---
https://bugs.gnucash.org/show_bug.cgi?id=797622

The Ninja-Build creates the default target at the bottom of `build.ninja` as follows:
```
#############################################
# Clean all the built files.

build clean: CLEAN


#############################################
# Print all primary targets available.

build help: HELP


#############################################
# Make the all target the default.

default all
```

As the target name help and directory name help/ conflict, `ninja clean` tries to remove the help/ directory.
The help/ directory usually contains files, then removing it always fails.

This PR is to avoid this problem.

The names of all, help, clean, install, default, and so on may have a potential problem.


jralls:
Sigh, ugly as heck but it does work. The test failures are because some of the autotools removal work has leaked in so `ninja distcheck` is failing.

FWIW this is Cmake's fault, not Ninja's. The Cmake Ninja Generator adds "help" to the build outputs list:
```
void cmGlobalNinjaGenerator::WriteTargetHelp(std::ostream& os)
{
  {
    cmNinjaRule rule("HELP");
    rule.Command = NinjaCmd() + " -t targets";
    rule.Description = "All primary targets available:";
    rule.Comment = "Rule for printing all primary targets available.";
    WriteRule(*this->RulesFileStream, rule);
  }
  {
    cmNinjaBuild build("HELP");
    build.Comment = "Print all primary targets available.";
    build.Outputs.push_back(this->NinjaOutputPath("help")); // <---- this line is what makes ninja clean want to remove the help directory
    WriteBuild(os, build);
  }
}
```
I didn't find any other cases where magic filenames are put on the outputs list.

usamimikamen-sk:
> FWIW this is Cmake's fault, not Ninja's. The Cmake Ninja Generator adds "help" to the build outputs list:

Thank you for correcting my mistakes.

You said it was ugly as heck, but I want a conclusion and a decision to write a README.md file.
I think there are two options.
(1) Stop using ninja for GnuCash-docs.
(2) rename help/ directory.

Do you have any opinions or other options?

jralls:
I think it was @gjanssens who said that it was a ninja bug.

Not using Ninja isn't an option with jhbuild. We can remove the clean command from the Windows build configuration so that it doesn't stop the build or we can change the name of the directory. Maybe we should do both.

gjanssens:
> I think it was @gjanssens who said that it was a ninja bug.
> 
I did indeed suggest that when I wrote bug 797622 as I found a few references to ninja not deleting directories with untracked files. However  I didn't discover the clash of cmake's help target and our  help document directory. 

It's an new issue we introduced ourselves by migrating from autotools to  cmake.

> Not using Ninja isn't an option with jhbuild. We can remove the clean command from the Windows build configuration so that it doesn't stop the build or we can change the name of the directory. Maybe we should do both.

I think only changing the name of the directory is sufficient. Stopping the build from using the clean command is just a stop-gap measure. Sooner or later someone will complain it doesn't work when they run it manually.

And to bikeshed a bit, I don't like `helpman`  as new name. I suggest either `manual` and `guide` or `gnc-help` and `gnc-guide`.

usamimikamen-sk:
@CWehli
Directory names guide/ and help/ will be changed in the future.

derekatkins:
@usamimikamen-sk --
>  Directory names guide/ and help/ will be changed in the future.

Any changes of this nature will also need to be coordinated so the build-server can be updated simultaneously.  The build server has "guide" and "help" directories hard-coded in the build script.

fellen:
> And to bikeshed a bit, I don't like `helpman` as new name. I suggest either `manual` and `guide` or `gnc-help` and `gnc-guide`.

Just `manual` and `guide`

usamimikamen-sk:
This PR consists of 3 steps to avoid conflicts.
Step1. modify CMakeLists.txt
Step.2 git mv help manual
Step.3 squash commits.

When you merge this PR into maint, do:
* cherry-pick Step 1,
* run Step 2 by your hands,
* squash if necessary.

PR #217, opened 2021-09-03
(5)remove autotools related data. (after 4.7 release?)

usamimikamen-sk:
depends on PR #215.
---
Remove autotools and related distribution settings.

This PR depends on PR #215.

---
 65dab67 rebase


derekatkins:
If (when?) this gets merged into maint, we need to coordinate because this *WILL* break the nightly builds on `code` which currently use the auto-tools build system.

gjanssens:
> If (when?) this gets merged into maint, we need to coordinate because this _WILL_ break the nightly builds on `code` which currently use the auto-tools build system.

You don't have to wait for this PR to get merged.  cmake support has been complete  for some time now. So  you  can switch to it beforehand. Unless our  nightlies run `make clean` (to become `ninja clean`) as that is the topic of PR #216 .

jralls:
> You don't have to wait for this PR to get merged. cmake support has been complete for some time now. So you can switch to it beforehand. Unless our nightlies run make clean (to become ninja clean) as that is the topic of PR #216 .

@derekatkins can use cmake with make (don't say `-G Ninja` in the cmake command) and avoid the ninja clean bug.

derekatkins:
@jralls --
> @derekatkins can use cmake with make (don't say `-G Ninja` in the cmake command) and avoid the ninja clean bug.

Currently the build server does not have cmake installed.  I can install cmake 3.14.5, provided that version works for us.

jralls:
@derekatkins 3.14 is fine. The minimum is 3.5 for gnucash-docs and 3.10 for gnucash itself.

cmake/AddGHelpTarget.cmake
```diff
@@ -77,6 +93,13 @@ function (add_ghelp_target docname lang entities figures dtd_files)
 
 
     install(DIRECTORY "${BUILD_DIR}"
-        DESTINATION "share/gnome/help/${docname}")
+        DESTINATION "share/help/${lang}")
+
+    # Cleaning
+    # ADDITIONAL_CLEAN_FILES is introduced in cmake 3.15 but we can't use it.
```
jralls, line 99: Please say instead "Can be replaced with ADDITIONAL_CLEAN_FILES once we can adopt Cmake 3.15."

jralls:
That last commit does a lot more than add an extra clean target. Please break it up into separate commits for each task and include in the commit message the reason for each. As you do that consider whether the change is really part of removing autotools or really belongs in a separate PR.

What's the point of cramming HACKING and CmakeNotes.txt into README.md?

usamimikamen-sk:
> That last commit does a lot more than add an extra clean target. Please break it up into separate commits for each task and include in the commit message the reason for each. As you do that consider whether the change is really part of removing autotools or really belongs in a separate PR.

I'm sorry. I will rearrange the branches and PRs.

> What's the point of cramming HACKING and CmakeNotes.txt into README.md?

Both HACKING and CmakeNotes.txt are outdated. I'd like to direct contributors to the Wiki.

cmake/AddGHelpTarget.cmake
usamimikamen-sk: I read the ADDITIONAL_CLEAN_FILES document and found they are not enough functionality. So I will change this code.

usamimikamen-sk:
I only picked up autotools removal and related changes.

jralls:
> Both HACKING and CmakeNotes.txt are outdated. I'd like to direct contributors to the Wiki.

If they're obsolete they should be deleted, not made more prominent in the file that displays under the directory listing on Github.

fellen:
I would prefer to drop autotools after the release.
Do you have parts you want to get added before?

jralls:
I don't see any rush to do any of this before the release.

PR #218, opened 2021-09-11
Remove some unresolved ID reference from Help-Files

CWehli:
* id="Using-Help"
* id="usage"


fellen:
I assume the purpose is to get rid of the warnings on pdf creation?

As they are not used in [gnucash/gnome-utils/gnc-ui.h](https://github.com/Gnucash/gnucash/blob/maint/gnucash/gnome-utils/gnc-ui.h),, we could remove them.

In both cases defining an id for an title, where already the section has one, makes not much sense.

Thanks!

CWehli:
Yes, that was the goal of this PR.

PR #222, opened 2021-09-24
Fix: Replace COPYING file from GPL 3 to GPL 2.

usamimikamen-sk:
The current GnuCash License is GPL 2.0 or later.
But gnucash-docs/COPYING is GPL 3.0.

This PR replaces COPYGIN with GPL 2.0.


fellen:
Merged, thanks!

PR #211, opened 2021-08-25
(1)remove OASIS xsl stylesheets from the tree and clean up old or unused customization files

usamimikamen-sk:
```
(1) restructure docbook xsl stylesheets.
  * The xsl/1.79.2/ was completely removed and gnucash-docs uses the system docbook-xsl stylesheets now.
  * The images of OASIS docbook-xsl was copied into xsl/images.
  * The admonition icons under stylesheets/ were merged into xsl/images/.
  * The stylesheets/ entry for distfiles in Makefile.am was removed.
  * The old GnuCash icons were updated and put into xsl/images/.
  * The XSL stylesheets for GnuCash customization were split.
       First, gnc-custom-chm.xsl, gnc-custom-epub.xsl, gnc-custom-fo.xsl
           and gnc-custom-html.xsl are for CHM, EPUB, FO/PDF and html, respectively.
       Then, gnc-custom-common.xsl is imported for common use.
  * Unused xls stylesheets were removed.
  * The variablelist.xsl was split into variablelist-fo.xsl and
    variablelist-html.xsl because they depend on formats.
  * The stylesheets for title page were cusomized.
  * xsl/README was updated.
  * For PDF:
     - PR #203 by @CWehli: The variable list customization was imported.
     - The paper size is set to USletter and A4 for C and the other locales respectively.
     - Other minor improvements were applied.
  * For CHM:
     - The Character code is set Shift_JIS and KOI8-R for ja and ru locale respectively.

(2) cmake related
  * The stylesheet/ was removed from distfiles in CMakeList.txt.
  * Usages comment of function was added at the top of the files.
  * AddGncDocTargets.cmake:
     - SVG files were removed from the list "${figure}" because they aren't used right now.
     - "${dtd_files}" was added to support multiple dtd files.
  * AddChmTarget.cmake:
     - MakeChm.cmake was merged.
     - The working directory was changed to "${CMAKE_CURRENT_BINARY_DIR}/htmlhelp".
     - The methods of copying figures and icons were changed to use add_custom_command.
     - Dependencies on "${dtd_files}" were added.
     - The dependencies were rearranged.
     - Unused functionalities in GnuCash, such as alias and map, were removed.
     - The "--stringparam gnc.lang ${lang}" argument is added to the xsltproc so as to swich the character codes.
     - The dtd file name for xsltproc was changed to support the new xsl/ structure.
   * AddEpubTarget.cmake:
     - The dtd file name for xsltproc was changed to support the new xsl/ structure.
[...]

PR #223, opened 2021-09-24
(4a) Set attribute of *list spacing="compact" for FO/PDF.

usamimikamen-sk:
Depends on #211
---
This is requested by @CWehli in https://github.com/Gnucash/gnucash-docs/pull/211#issuecomment-923815297.
This PR is for FO/PDF only.

I'll create another PR for HTML(s) because it is a bit complex to do.
It becomes clear by using external CSS. So I have to set some parameters, create CSS file, and change cmake files.


usamimikamen-sk:
unspecified
compact

CWehli:
@usamimikamen-sk 
Thanks.
> This PR is for FO/PDF only.

Do you have the possibility to underline the links in fo/pdf? (like a hyperlink in HTML)

usamimikamen-sk:
> Do you have the possibility to underline the links in fo/pdf? (like a hyperlink in HTML)

@CWehli 

There are three types of links in PDF.
(1) `See 1.1 title` (xref title)
(2) table of contents
(3) `The Free Software Foundation [http://www.gnu.org/fsf/fsf.html] ` (ulink external URL in the brackets)

Which links?

CWehli:
@usamimikamen-sk 
> (1) `See 1.1 title` (xref title)
> (2) table of contents
> (3) `The Free Software Foundation [http://www.gnu.org/fsf/fsf.html] `

I think for numbers 1 and 3 the underlining makes sense.
In TOC, the link must work, but I don't like the underlining there.

fellen:
From my understanding in pdf because of its paper origin it is more common to append square brackets with the links.

CWehli:
>  square brackets with the links.

This is also ok for me, but it is important that such links are visually highlighted.

fellen:
> [Edit] I think there is also a bug entry for this. [/Edit]

[Bug 796854 - Internal document links are not visually discoverable in PDF - need separate styling](https://bugs.gnucash.org/show_bug.cgi?id=796854)

PR #225, opened 2021-09-24
Guide:C: Add a directory with datafiles for faster regeneration of images

fellen:
This commit was split from commit 014ce27 to avoid adding new files before all the PRs with structural changes on the docs.

Reason to add the directory:
Each time you want to regenerate an image in the guide because something changed in the appearance of the program, you have to read and replay the chapter in the program to get the state and take the screenshot. By storing the datafiles here, we can save much time.

Is it sufficient to have them in the repo or should they also go in the tarball?

### DIFF
--- a/data/C/guide/currency_main1.gnucash
+++ b/data/C/guide/currency_main1.gnucash
@@ -1,0 +1,352 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<gnc-v2
+     xmlns:gnc="http://www.gnucash.org/XML/gnc"
+     xmlns:act="http://www.gnucash.org/XML/act"
+     xmlns:book="http://www.gnucash.org/XML/book"
+     xmlns:cd="http://www.gnucash.org/XML/cd"
+     xmlns:cmdty="http://www.gnucash.org/XML/cmdty"
+     xmlns:price="http://www.gnucash.org/XML/price"
+     xmlns:slot="http://www.gnucash.org/XML/slot"
+     xmlns:split="http://www.gnucash.org/XML/split"
+     xmlns:sx="http://www.gnucash.org/XML/sx"
+     xmlns:trn="http://www.gnucash.org/XML/trn"
+     xmlns:ts="http://www.gnucash.org/XML/ts"
+     xmlns:fs="http://www.gnucash.org/XML/fs"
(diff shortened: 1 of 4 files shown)


jralls:
> Is it sufficient to have them in the repo or should they also go in the tarball?

Unless they're needed for generating the released docs I don't they need to be in the tarball.

Do you have a plan to create a set for each language?

fellen:
> Do you have a plan to create a set for each language?

I thought, I leave it to the translators.

jralls:
That might be asking too much, but maybe not. Regardless I'd create the directories in all the languages.

What about help? It doesn't have the same story-line progression of illustrations but it does have illustrations and ISTM the same book could be used to generate them. In that case it would make sense to have the datafiles directory at root level with its own language subdirectories instead of in each language under `guide`.

fellen:
@CWehli Would it make sense for Help illustrators or could you reuse that from guide?
If you want your own directory, you estimation of required files and their structure?

@jralls Currently we have the rule  for new translations: Copy C and translate it.
That needs a review after XDG adaption.

CWehli:
I like the idea of providing a ready-made data file per language for the help documentation and also some import files (csv, qif, etc.).
For structuring purposes, a common directory per language with the example files at the top level seems to me to make sense.

gjanssens:
Will there be more than one book per language ? If not, it may be enough to have a single datafiles directory and name the books C.gnucash, de.gnucash,...

Do you expect to have example qif and csv files translated as well ?

I'm just trying to establish the need for additional language directories in the datafiles directory or not. I'm fine with both though.

CWehli:
For the help documentation, one data file per language is probably sufficient.

The import files should also be translated so that the screenshots are comprehensible throughout and can be described more easily in the text.

fellen:
> Will there be more than one book per language ?

Yes, for the guide at least one serie of books per chapter.
Having the entries from mortgage in the same book as the entries currency might confuse the user or becomes very hard to take proper screenshots–strange  content in the summary bar.

> If not, it may be enough to have a single datafiles directory and name the books C.gnucash, de.gnucash,...

…and this names are then displayed in the window's title bar.

With the XDG adaption in mind I would suggest
datafiles/LANGUAGE/COMPONENT[/SUBCHAPTER]
with COMPONENT: {manual | guide [|common] }

PR #226, opened 2021-09-25
Workaround patch for install gnc-locale-${lang}.dtd for CMake.

usamimikamen-sk:
Conflict PR #211
---
This patch is for PR #207.

cmake -S gnucash-docs -B .
cmake --build . --target ghelp
cmake --install . --prefix ${prefix}

Then ghelp will be installed under ${prefix}/share/gnome/help/ with correct dtd files.


fellen:
Just in time, thanks!

PR #227, opened 2021-09-25
(5b) Remove ScrollKeeper's OMF file.

usamimikamen-sk:
Depends on PR#207
---
ScrollKeeper is deprecated, and it seems not to be used anymore. But omf files exist in each document directory.

This PR removes OMF files.

The entries of OMF files exist in both CMake and autotools. So I recommend applying after PR#207.


fellen:
Why can you not base it on Maint? 
That would perhaps reduce the number of files in the XDG adjustment.

usamimikamen-sk:
> Why can you not base it on Maint?
> That would perhaps reduce the number of files in the XDG adjustment.

I think this PR is not urgent. I don't think this PR is not necessary to include release 4.7. 
If you want to include this PR in 4.7, I will rebase against maint.
Another reason is that we are in a hurry for PR # 228.

jralls:
We're not in a hurry for any of this, and as I commented on another PR I don't want anything merged this weekend. 

@fellen Please stop asking to modify autotools, that's a waste of effort.

@usamimikamen-sk I understand that this work has proceeded somewhat organically, but there's no interdependence between removing autotools, the xsl stylesheets, and the scrollkeeper files. Had you made those changes in separate feature branches all based on a maint commit then they could be addressed independently and the obvious ones--like this one--merged immediately while the more difficult or controversial ones could continue to be hashed out.

PR #228, opened 2021-09-25
Change to use TTF instead of OTF for Russian PDF files.

usamimikamen-sk:
Recent fop fails to read OTF files to embed into PDF. Therefore TTF shall be used.

TODO: 
* Change gnucash-docs/fonts directory structure.
* Organize license files.

### DIFF
--- a/cmake/AddGHelpTarget.cmake
+++ b/cmake/AddGHelpTarget.cmake
@@ -18,7 +18,13 @@ function (add_ghelp_target docname lang entities figures)
     foreach(xml_file ${entities} ${docname}.xml)
         list(APPEND source_files "${CMAKE_CURRENT_SOURCE_DIR}/${xml_file}")
     endforeach()
-    list(APPEND source_files "${CMAKE_SOURCE_DIR}/docbook/gnc-docbookx.dtd")
+
+    set(dtd_files "${CMAKE_SOURCE_DIR}/docbook/gnc-docbookx.dtd"
+                  "${CMAKE_SOURCE_DIR}/docbook/gnc-locale-C.dtd"
+                  "${CMAKE_SOURCE_DIR}/docbook/gnc-locale-${lang}.dtd")
+    list(REMOVE_DUPLICATES dtd_files)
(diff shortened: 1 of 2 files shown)


fellen:
We have still autotools. So one change is missing

fellen:
… and they should then be under the truetype directory instead of opentype, I assume.
@jralls or were there other reasons for the current layout?

fellen:
@jralls I got it working for autotools.
Should we drop freetype openfonts in favour of their corresponding TTFs?

jralls:
Let's stop this right now, it's stupid. The Russian translation effort failed after translating only the Overview chapter and is unfit for distribution: Distributing a claimed translation that translates only a few pages makes us look ridiculous, it's even worse than the Italian and Japanese translations that haven't been maintained for 10 years and are badly outdated.

@fellen, your hope that someone would take up the Russian translation has come to naught and now you've led @usamimikamen-sk to waste time finding fonts for it.

usamimikamen-sk:
```
yasuaki@debweb01:~/gcdev/usamimikamen_sk/b-my$ make de-gnucash-help-pdf
[ 33%] Generating gnucash-help.fo
Making portrait pages on USletter paper (8.5inx11in)
[ 66%] Generating ../../share/doc/de/gnucash-help.pdf
(snip)
[WARN] FOUserAgent - Glyph "?" (0x130, Idot) not available in font "Times-Roman".
(snip)
[100%] Built target de-gnucash-help-pdf
```

In German PDF, a glyph of U+0130 Idot is not available. So I think it is better to specify Unicode fonts for German documents.

I did `reset --hard` in the maint branch of my forked repository. 
Therefore GitHub closed this PR automatically.

The changes can be seen in bd486f6.

jralls:
It would be better to specify Unicode-compliant fonts for all documents, but that shouldn't mean we need to include the fonts in the repo. We don't need to make PDF generation cross-platform, Linux is good enough. 

> I did reset --hard in the maint branch of my forked repository.

Oops. That won't happen if you do pull requests from their own feature branches.

PR #229, opened 2021-09-26
(1)Remove autotools and some unused files.

usamimikamen-sk:
I think autotools should be removed first because maintenance costs are large.

Other unused or obsolete files should be removed.
* ScrollKeeper's OMF files.
* gnucash-docs.spec.in.
* CMakeNotes.txt
* HACKING


jralls:
Good start, but 3 commits please: 1 each for autotools, OMF files, and other stuff.

gjanssens:
Ah, what a relief to finally see these go. And I agree completely it makes sense to drop autotools before anything else. It saves on refactoring work for your other PRs.

gjanssens:
Thanks!

CMakeLists.txt
```diff
@@ -196,7 +196,6 @@ add_to_dist(
     COPYING
     COPYING-DOCS
     ChangeLog
-    HACKING
```
fellen, line 218: This file is still mentioned in README.

Tip: Before removing files run a command like `grep -inr 'HACKING'` in the top source directory.

gjanssens: Thanks for the heads up @fellen . I assume you will handle this trivial fix yourself ?

PR #230, opened 2021-09-26
(1a) Rename directory name help/ to manual/ in order to fix Bug#797622.

usamimikamen-sk:
Depends on #229
---
https://bugs.gnucash.org/show_bug.cgi?id=797622

The Ninja-Build creates the default target at the bottom of `build.ninja` as follows:
```
#############################################
# Clean all the built files.

build clean: CLEAN


#############################################
# Print all primary targets available.

build help: HELP


#############################################
# Make the all target the default.

default all
```

As the target name help and directory name help/ conflict, `ninja clean` tries to remove the help/ directory.
The help/ directory usually contains files, then removing it always fails.

This PR is to avoid this problem.

The names of all, help, clean, install, default, and so on may have a potential problem. 
---
This PR consists of 3 steps to avoid conflicts.
Step1. modify CMakeLists.txt
Step.2 git mv help manual
Step.3 squash commits.

When you merge this PR into maint, do:
* cherry-pick Step 1,  8ac3328
* run Step 2 by your hands, `git mv help manual`
* run Step 3 by your hands, `git commit --amend`


jralls:
There are some `Makefile.am` that seem to have escaped #229 and the distcheck CI s are failing because of something to do with autogen.sh, also suggesting that #229 isn't quite complete.

usamimikamen-sk:
> There are some `Makefile.am` that seem to have escaped #229 and the distcheck CI s are failing because of something to do with autogen.sh, also suggesting that #229 isn't quite complete.

This PR is created against maint, not rm-atools3. So distcheck failed.
If PR #229 is applied, This PR will work fine.

usamimikamen-sk:
This PR can apply anytime after PR #229.

gjanssens:
Ok, though contrary to @jralls '  comment on #229 this really should only be one commit. The first commit in this PR wouldn't build without the second one even after #229 is merged.

The idea is that each commit should be atomic and "small enough" to be reviewable. That 'small enough' is vague and takes some time to master. The changes in the two commits in the PR together are easy to understand: you move a bunch of files around and update CMakeLists.txt accordingly. While it's a lot of files that get moved around, it's still part of one single change: moving files without changing CMakeLists.txt or the other way around both would result in an intermediate commit which breaks the documentation build.

#229 was a counterexample: you did 3 things in one commit that could each be done separately: drop autotools, drop obsolete omf and then a remaining cleanup. Putting all of that in one big commit blurs the lines between what change was for what reason. Hence it would afterwards be more difficult to go bug hunting.

Should you wonder why that is, git has a feature called bisect that helps trace a bug to a very specific commit. This can only work well if all commits focus on one atomic change (to reduce the amount of interpretation needed by the reviewer) and each commit it itself builds successfully.

usamimikamen-sk:
@gjanssens
Thank you for your advice.

But I can't expect when this PR is merged into maint.
This PR needs coordination with @derekatkins to work on build-server.
And whenever files under help/ are changed,  conflicts always occur in 9eac75c.
So I proposed to use three steps.

Step 3 may be better to use `git commit --amend`.

gjanssens:
> @gjanssens
> Thank you for your advice.
> 
> But I can't expect when this PR is merged into maint.
> This PR needs coordination with @derekatkins to work on build-server.

Duly noted :)
My plan was to coordinate this with @derekatkins at the time of merging this PR. For the record, not only the build server will need to be checked but also our packaging scripts for our Windows, flatpak and  Macos installers. And likely also our website code (gnucash-htdocs and gnucash-htdocs-docs repositories).

> And whenever files under help/ are changed, conflicts always occur in [9eac75c](https://github.com/Gnucash/gnucash-docs/commit/9eac75c4bd399e1364d5722c9aa23e248d09ac38).
> So I proposed to use three steps.

Ok, leave it as two commits then  for now. We'll see how to handle it at merge time.

jralls:
If @derekatkins hasn't already switched the nightly build to cmake then merging this will motivate him. ;-)

gjanssens:
> If @derekatkins hasn't already switched the nightly build to cmake then merging this will motivate him. ;-)

I did  work with @derekatkins a few days back to get the nightly build migrated to cmake. My worry was about the source directory name change from `help`  to `manual`. Turns out there was  no need for that. None of our packaging repos directly interact with the source level directory names. They all just to a top-level `make/ninja <something>` with that `<something>` being html or nothing. The resulting files are copied from the build or install directory. And the build results we do care about (the html, pdf, epub, chm or ghelp files) are stored in common directories using gnucash-help or gnucash-docs as intermediate directory not help/manual. So none of the scripts in the other repos or on the build server need further tweaking.

Having said that, merging this commit may cause some concern for @CWehli as the help source files move to a new location.  So some of his PRs will run into merge conflicts. There are a few now already because of changes to now removed autotools files. There will be more due to new files being added in help that should now go to manual.

So question is, when do we merge this one ?
@CWehli do you think you can handle the merge conflicts or should we wait for your PRs touching help to be merged first ?

gjanssens:
Note this PR itself also has merge conflicts due to already removed autotools/omf files being moved. I was able to handle those so there's not really a need to do more for this PR per se.

CWehli:
> Having said that, merging this commit may cause some concern for @CWehli as the help source files move to a new location. So some of his PRs will run into merge conflicts. There are a few now already because of changes to now removed autotools files. There will be more due to new files being added in help that should now go to manual.
> 
> So question is, when do we merge this one ?
> @CWehli do you think you can handle the merge conflicts or should we wait for your PRs touching help to be merged first ?

I'm not well-versed in fixing merge conflicts and using the git tools.
My procedure at this point is to cache my edited files outside the directory structure, rebase them and copy the saved files back in.

However, I think I can fix the conflicts.

gjanssens:
That's good. This morning I test-merged your PR #206 on  top of this PR. I have no issues solving the conflicts due to files moving around,  but there also content  conflicts of which I don't know exactly  what end result you would like (likely because these files have been changed on the maint branch after you  started #206 ). But if that's something you can deal with, I will merge this bit in soon.

CWehli:
@gjanssens 
Yes, I'm already looking at it and getting the merge conflicts corrected again as well.

gjanssens:
Ok, then. I have squashed the commits, updated to commit message to explain why this move is needed and pushed the result to maint. Thanks for finding and fixing this  @usamimikamen-sk

fellen:
I had hoped,it were renamed both in source and target. "Help" is a really stupid name. It was right as long as there was only one component. But now there are a guide and a manual and both are available by the menu "Help".

PR #246, opened 2022-01-09
FO formatting, Added parameter to PdfTarget...

CWehli:
to improve nesting of 'variablelist' elements.
Some users have criticised the presentation in the PDF documentation.


fellen:
Thanks!
My comment is in commit 27a39cf.

PR #250, opened 2022-01-24
[Bug 798414] Help_ch_GUIMenues: How do I resize my register columns?

CWehli:


CWehli:
@jralls: Hi, please take a look at this PR.

jralls:
@CWehli This is mostly documented in [the Guide](https://www.gnucash.org/docs/v4/C/gnucash-guide/chapter_txns.html#txns-columnwidths1"), [code](https://github.com/Gnucash/gnucash-docs/blob/maint/guide/C/ch_txns.xml#L197). The missing piece is about changing the split columns when in split view.

Arguably the instructions about manipulating the register belong in the Help with pointers in the Guide rather than the other way around, but we should leave it the way it is for fixing the 4.x docs. Rather than creating a separate tip let's add another pointer to the one about selecting the register style, something like
```
  <tip>
        <para>Appearance of the Account Register Display is highly customizable (see
          <ulink
url="https://www.gnucash.org/docs/v&series-stable;/C/gnucash-guide/chapter_txns.html#txns-regstyle1">
          Tutorial and Concepts Guide, Choosing a Register Style</ulink> and <ulink url="https://www.gnucash.org/docs/v4/C/gnucash-guide/chapter_txns.html#txns-columnwidths1">Setting Column Widths"</ulink>).
        </para>
      </tip>
```
Then add the bit about needing to put the focus in a split to adjust the split columns to the Setting Column Widths section in the Guide. I think that a screen shot showing the header change with a split selected would be helpful.

Another note, @fellen might disagree: I don't think we want pointers to the Wiki in the documentation, for two reasons: First, the wiki gets changed a lot and it's really easy for a link to it in the docs to go stale. Second, the conceptual flow of documentation is from the wiki to the Help and Guide, so if there's material in the wiki that we want added to the formal docs then we should do that and put a link to the docs in the wiki instead of linking the wiki from the docs.

CWehli:
> Then add the bit about needing to put the focus in a split to adjust the split columns to the Setting Column Widths section in the Guide. I think that a screen shot showing the header change with a split selected would be helpful

I don't want to insert a picture, because on the one hand I lack the database and on the other hand I use a different desktop theme. That would change the look of the guide a lot.

guide/C/ch_txns.xml
```diff
@@ -213,7 +213,13 @@
...
           though, as there's another column called rate whose width is set to 1 pixel to hide it
           (it's for internal use). Approach the divider from the left side and click as soon as the
-          cursor changes to a double-headed arrow.
+          cursor changes to a double-headed arrow (&lt;-|-&gt;).
+        </para>
+      </note>
+      
+      <note>
+        <para>The column headings are different for transactions and splits. Please keep this in mind if you want
+          to adjust the column width.
```
jralls, line 222: I think we need a little bit more than that. How about
```
When in split mode there are different column markers in the header for split rows and transaction rows depending on which type of row is active. In order to adjust split column widths one must first make a split row active; to adjust transaction line column widths a transaction row must be active.
```

jralls (changes requested):

jralls:
LGTM. @fellen?

manual/C/Help_ch_GUIMenus.xml
```diff
@@ -3264,7 +3264,9 @@
         <para>Appearance of the Account Register Display is highly customizable (see
           <ulink
 url="https://www.gnucash.org/docs/v&series-stable;/C/gnucash-guide/chapter_txns.html#txns-regstyle1">
-          Tutorial and Concepts Guide, Choosing a Register Style</ulink>).
+          Tutorial and Concepts Guide, Choosing a Register Style</ulink>)  and <ulink
+url="https://www.gnucash.org/docs/v&series-stable;/C/gnucash-guide/chapter_txns.html#txns-columnwidths1">Setting
```
fellen, line 3268: Please use the entity &url-docs-build-C;

@jralls For now the crosslinks go to the build server because that are more recent. OTOH code has more downtime than www.
Or we expand the link to something like `<ulink … code/…><footnote>If that fails, try <ulink … www/…></footnote>`

fellen:
> Another note, @fellen might disagree: I don't think we want pointers to the Wiki in the documentation, for two reasons: First, the wiki gets changed a lot and it's really easy for a link to it in the docs to go stale. Second, the conceptual flow of documentation is from the wiki to the Help and Guide, so if there's material in the wiki that we want added to the formal docs then we should do that and put a link to the docs in the wiki instead of linking the wiki from the docs.

There are more reasons: I was told the [Great Firewall](https://en.wikipedia.org/wiki/Great_Firewall) denies access to wiki pages.

@CWehli <q>The conceptual flow of documentation …</q> should be mentioned in "Getting Help" and its Guide counterpart.

manual/C/Help_ch_GUIMenus.xml
CWehli: I prefer to link to the official documentation in the www according to the release.
Since the versions are released every three months, the timeliness of the documentation is not crucial.

guide/C/ch_txns.xml
```diff
@@ -213,7 +213,14 @@
           it larger. Just drag its right-hand border out of the window to grow it. Be careful,
           though, as there's another column called rate whose width is set to 1 pixel to hide it
           (it's for internal use). Approach the divider from the left side and click as soon as the
-          cursor changes to a double-headed arrow.
+          cursor changes to a double-headed arrow (&lt;-|-&gt;).
```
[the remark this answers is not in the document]
CWehli: Since the double arrow is rarely needed, I think a separate entitiy for it is overkill.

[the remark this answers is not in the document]
fellen: Partially right. It has nothing to localize, so it belongs into gnc-docbookx.dtd. At some point someone might find a more appropriate enity in https://tdg.docbook.org/tdg/4.5/ref-charents.html, perhaps hArr  or harr from [ISOAMSA](https://www.w3.org/2003/entities/iso8879doc/isoamsa.html) and then we have only one place to change it for all languages.

manual/C/Help_ch_GUIMenus.xml
fellen: I don't object to change it—after the release.

guide/C/ch_txns.xml
```diff
@@ -213,7 +213,14 @@
...
+        </para>
+      </note>
+      
+      <note>
```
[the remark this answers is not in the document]
fellen: I meant the second note, not the first.

manual/C/Help_ch_GUIMenus.xml
fellen: The release manager can now swwitch ENTITY url-docs  from "&url-docs-draft;" to "&url-docs-release;"
https://wiki.gnucash.org/wiki/index.php?title=Documentation_Release_Process&type=revision&diff=21109&oldid=17299

PR #251, opened 2022-02-11
Help: generic transaction import matcher: Document new append checkbox

goodvibes2:
Commits:
1) fix adjust-dpi.sh so it does not update some files but not actually do anything
2) Update Preferences documentation as I want to refer to something on the undocumented Import tab
3) Add some minor improvements to Gen Imp Trans Matcher
4) Update Gen Imp Trans Matcher re new Append checkbox
5) Fix 2 warnings when building help pdf:
[WARN] FOUserAgent - Destination: Unresolved ID reference


goodvibes2:
Ping. String freeze 13/3/2022.

fellen:
> Ping. String freeze 13/3/2022.

That is only valid for the program code, neither docs nor htdocs.

util/adjust-dpi.sh
```diff
@@ -22,19 +22,20 @@ do
...
   #  and/or do not round the return value of identify -format "%x"
-  existing_dpi=$( echo "$existing_dpi" | awk '{ printf("%.1f",$1); }')
+  existing_dpi=$( echo "$existing_dpi" | awk '{ printf("%.2f",$1); }')
   # set the future dpi figure to XX.XX (PixelsPerCentimeter)
   future_dpi="$dpi_cm"
-  # apply new dpi only if it's changed from the existing
-  if [ "$existing_dpi" != "$future_dpi" ]; then
+  # apply new dpi only if it's changed from the existing by more than 0.02
+  abs_diffx100=$( echo "scale=4; ($existing_dpi - $future_dpi) * 100" | bc | tr -d - | awk '{ printf("%.0f",$1); }')
+  if [ "$abs_diffx100" -gt 2 ]; then	# really .02
```
jralls, line 36: This is silly. 

Image density should be 300/inch | 188/cm: That will provide a nice clear image on the highest-end displays and is what consumer printers are optimized for. At 188/cm, a 0.02/cm change would require a 1/2-meter wide image to change the number of pixels by 1: There is no such thing as a fractional pixel. Even at the 34 p/cm of a 1990-era 600x800 12" monitor it would take more than twice the width of the screen for a .02p/cm change to make a 1 pixel difference.

There's a comment at line 16 that says round to 0 decimals, that is integer p/cm. That seems reasonable for rounding, but for deciding whether to actually change an image if it's already within 5-10% of the target size, IOW 170-205 p/cm, it's good enough and should be left alone.

jralls (changes requested):

manual/C/Help_ch_GettingHelp.xml
```diff
@@ -308,4 +308,17 @@
...
+    <para>This document refers to mouse buttons as Button1, Button2 etc.</para>
+      <itemizedlist>
+        <listitem>
+          <simpara>Button1 is typically the left mouse button or the right button in a left-handed setup</simpara>
+        </listitem>
+        <listitem>
+          <simpara>Button2 is typically the right mouse button or the left button in a left-handed setup</simpara>
+        </listitem>
+      </itemizedlist>
+  </sect1>
```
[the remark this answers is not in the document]
goodvibes2: Hi John,
"Primary click" and "Secondary click" makes sense to me but I think still needs some explanation for others. Would that be OK?

util/adjust-dpi.sh
goodvibes2: Hi John,
Geert figured out what the dpi's should be. It seems to me we don't want the images being possibly 5-10% larger that what fits on the printed page, so I suggest we only adjust the image if it is more than an integer no of pixels/cm different from required. That gets rid of the fudge for doing non integer maths in bash. OK?

manual/C/Help_ch_GettingHelp.xml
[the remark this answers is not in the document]
jralls: Sure, just find a way to explain it that makes sense to somebody who's never seen a mouse.

util/adjust-dpi.sh
jralls: @goodvibes2 Well, it at least gets back to the intent of your line 16 comment, but I think it's still going to create unnecessary work. 

We may be approaching the problem from the wrong direction. Dockbook provides several parameters for controlling how an [image is displayed](http://www.sagehill.net/docbookxsl/ImageSizing.html) on the page. It's not as expressive as CSS but it does afford the ability to set a viewport at 100% of the page width (after margins) and to have the image scaled to fit in that space. Using those would ensure that the layout was what we want and then we'd need worry only about providing images that scale well on modern monitors and printers. Unfortunately Docbook hasn't really made it into the 21st century and doesn't support multiple-resolution images for different display conditions, so we will have to work around that by providing high resolution images and hope that they scale down decently on normal-res monitors; It's well known that scaling up a normal-resolution image to HiDPI doesn't look good.

@gjanssens, any thoughts? And where did the original DPI values come from?

jralls:
I've cherry-picked in all the commits except the first, "Update adjust-dpi.sh so rerun does nothing" so the docs for the new append checkbox are in tomorrow's release.

fellen (commented):
Sorry, I ust found out, I had not pushed ny review. It is now more info of other changes in between.

@goodvibes2 can you pick up the warning about bayes learning?

util/adjust-dpi.sh
fellen: > Image density should be 300/inch | 188/cm: That will provide a nice clear image on the highest-end displays and is what consumer printers are optimized for.

That is not relevant. `<imageobject role="html">` is scaled by `<imagedata … width=…>`
Only `<imageobject role="fo">` used for pdf relies on the images dpi property for scaling.
> @gjanssens, any thoughts? And where did the original DPI values come from?

I would assume the defaults for PDF generation.

jralls: @fellen, this isn't about the scaling that's passed to docbook or even embedded in the image tags. This is about the scaling of the png files that are scaled for presentation. The right way to do this for a web page is to have several copies of an image at different resolutions have have server-side code that asks the browser what it needs and serves that resolution. We can't do that, because neither docbook nor our hand-rolled php website support it so the only thing we can do is serve high-dpi images and let the browser downsample them for display on the few remaining low-resolution monitors in the wild. The current low res documentation images get upsampled instead and look fuzzy on a high-DPI monitor.

We don't care too much about color fidelity in these images so to keep the page load time from being ridiculous we can crank the compression way up.

fellen: > The current low res documentation images get upsampled instead and look fuzzy on a high-DPI monitor.

Can you name an example?

jralls: Here's an example from the Guide. Compare the text in the image to the html text:
Screenshot 2022-08-26 at 7 47 47 PM

The element in question is
```
<div class="screenshot"><a name="txns-registersplit4"></a><div class="mediaobject"><img src="[figures/txns_registersplit4.png](view-source:https://www.gnucash.org/docs/v4/C/gnucash-guide/figures/txns_registersplit4.png)" alt="3 Jeans purchases">
```

And since you labor under the illusion that you've somehow fixed it in the manual, here's one from there:
Screenshot 2022-08-26 at 7 59 59 PM
Instead of fuzzy it's so small that it's useless. The element:
```
<img src="[figures/Help_Tool_LotsStk.png](view-source:https://www.gnucash.org/docs/v4/C/gnucash-help/figures/Help_Tool_LotsStk.png)" width="510" alt="Example of Lots in Account window for a security account">
```
Note the hard coded width in the default pixels. *That's* wrongly specified, not the guide's. The monitor I'm using is 3024x1964 physical pixels, 1512x982 logical ones. That help (since it's v4) image is 9x3.5 cm. Yeah, I can zoom in on it. You know what that makes it? Yeah, fuzzy.

I have screens ranging in physical size from 6x12 cm (iPhone 13 mini) to 60x30 (LG 4K 27" monitor). Those two plus the main laptop are all around 100 pixels/cm. I have another non-HiDPI 27" monitor that's the same size as the LG but has about 1/2 the resolution. Modern web pages have to be made to look good on all of them. Fixed pixel scaling is *not* the way to get there.

fellen: > And since you labor under the illusion that you've somehow fixed it in the manual, here's one from there: Screenshot 2022-08-26 at 7 59 59 PM Instead of fuzzy it's so small that it's useless. The element:
> 
> ```
> Example of Lots in Account window for a security account
> ```
> 
> Note the hard coded width in the default pixels. 

That is my recent work: replace width="510px" by &img-w;, which is currently defined as "800px". That again seems to be the default width of yelp.

> _That's_ wrongly specified, not the guide's. The monitor I'm using is 3024x1964 physical pixels, 1512x982 logical ones. That help (since it's v4) image is 9x3.5 cm. 

This changed with commit 1ff3733 to
grafik
You have to use https://code.gnucash.org/docs/C/gnucash-help/tool-lots.html#idm45337239301024 for recent changes. In https://www.gnucash.org/docs/v4/ it will appear with your next release.
> Yeah, I can zoom in on it. You know what that makes it? Yeah, fuzzy.

Try right-click-><q>Open graphic in separate tab</q> and you see the original size.
Perhaps we should add this tip to the to be written conventions of the docs.

fellen: > Here's an example from the Guide. Compare the text in the image to the html text: Screenshot 2022-08-26 at 7 47 47 PM
> 
> The element in question is
> 
> ```
> <div class="screenshot"><a name="txns-registersplit4"></a><div class="mediaobject">3 Jeans purchases
> ```

This is a very old image. I assume, when Bengt rewrote that part he didn't know how to use scaling and shrinked his images to 510px width.

Obvisious we should regenerate his images. For now I filed [bug 798608](https://bugs.gnucash.org/show_bug.cgi?id=798608) for this issue.

jralls: >You have to use https://code.gnucash.org/docs/C/gnucash-help/tool-lots.html#idm45337239301024 for recent changes. In https://www.gnucash.org/docs/v4/ it will appear with your next release.

Yes, that looks better, though it's still using a fixed width instead of a percentage, so it's about 60% of the laptop window (half if I put the browser in full-screen) and 80% on the phone (which is of course always full-screen).  Interestingly the phone browser scales everything, text included, to manage that same width while the laptop browsers don't. That makes the text tiny if the phone is held upright; turning it sideways maintains the same width proportion and makes the text large enough to read without a magnifier.  Meanwhile on the laptop having the text span the whole 30cm of the monitor (never mind the 60cm of the 27" desktop) is uncomfortable to read. [W3C recommended practice](https://www.w3.org/WAI/tutorials/page-structure/styling/#line-length) is 80 characters/column. They suggest setting `max-width` to 70em or allowing the user to adjust the width of the page. Our design does the latter, but because you hard-scaled the image and its (completely unnecessary) containing table at 800px the image won't shrink with the window. Scaling it to 80% instead would allow it to adjust to the page size. OTOH I think setting a `max-width` on the page would improve readability for everyone. That wouldn't make it OK to hard-code image pixel widths: Users might still want to narrow the page or have a monitor that makes that necessary and the images should adapt to that resizing.

>>    Yeah, I can zoom in on it. You know what that makes it? Yeah, fuzzy.

> Try right-click->"Open graphic in separate tab" and you see the original size.
Perhaps we should add this tip to the to be written conventions of the docs.

More usefully, I downloaded it so that I could get its true size and resolution: 1680 x 651 px at ***305 px/in***.
Exactly what I recommended on March 10 and what you objected to. Thank you for proving my point.

PR #259, opened 2022-04-24
dtd-docbook: Added an explanation of the syntax and corrected some previous entries.

CWehli:


docbook/gnc-locale-de.dtd
```diff
@@ -27,46 +27,46 @@
...
 
 <!-- Window-Menu -->
 <!ENTITY % gm.win "<guimenu><accel>F</accel>enster</guimenu>">
-<!ENTITY % gmi.nwin "<guimenuitem><accel>N</accel>eues Fenster</guimenuitem>">
-<!ENTITY % gmi.npage "<guimenuitem>Neues Fenster mit <accel>S</accel>eite</guimenuitem>">
+<!ENTITY % gmi.win.win "<guimenuitem><accel>N</accel>eues Fenster</guimenuitem>">
+<!ENTITY % gmi.win.page "<guimenuitem>Neues Fenster mit <accel>S</accel>eite</guimenuitem>">
 
 <!-- Help-Menu -->
 <!ENTITY % gm.help "<guimenu><accel>H</accel>ilfe</guimenu>">
```
fellen, line 65: How about `gm.hlp`?
`grep -inr 'gm.help' .` returns only the dtds. For the entries below
`grep -inrF '.help.' .` returns also /manual/de/Help_ch_GUIMenus.xml

docbook/gnc-locale-C.dtd
```diff
@@ -1,8 +1,6 @@
 <!-- Add master English entities that should be translated below.
      Only parameter entities (% entities) should be translated.
-     See htps://www.w3.org/TR/REC-xml/#sec-physical-struct. -->
-
-
+     See https://www.w3.org/TR/REC-xml/#sec-physical-struct. -->
```
fellen, line 3: should we add something like
As an example compare this file with the german localization gnc-locale-de.dtd.

Perhaps we should explain it better for "simple" translators:

Motivation:
With this magic we have only one place to maintain menu entries and it is consistent for the user in the whole document—without typos.

Task:
Start Gnucash in your language and update this list to your menu entries.

As the list is overwelming on the first view: Which part is finally used in the docs?

fellen (commented):
Amost finished! ;-)

docbook/gnc-locale-de.dtd
CWehli: Correct, this was only meant to be a test to see if it would work. The actual translation and further work on the individual chapters will come with a later commit.

docbook/gnc-locale-C.dtd
CWehli: > As the list is overwelming on the first view: 

This file is not really clear, maybe we can think of something else to structure it? I think if we can split the file even further, one with just the GUI entries for translation and another with the entity combinations for use, then this can become more manageable.

> Which part is finally used in the docs?

In GUIMenus.xml most of the individual parts are used. The entities for use in the respective chapters are still to be declared as menuchoice so that they can be used in the text.

docbook/gnc-locale-C.dtd
```diff
@@ -39,171 +53,555 @@
 
 <!-- guimenu/guisubmenu/guimenuitem to be used as names in text -->
 <!-- <!ENTITY % gmi.xxx "<guimenuitem><accel></accel></guimenuitem>"> -->
-<!-- <!ENTITY gmi. "%gmi.;"> -->
+<!-- <!ENTITY gmi.xxx.yy-zz "%gmi.xxx.yy-zz;"> -->
+<!-- Description: 
+     gm:  Entry in the menubar
+     gmi: Every menu entry, also in submenus
```
fellen, line 59: I wonder why menu entry get's abbreviated by gmi? Its derived from `GuiMenuItem?
Perhaps we should use such terms then here too.

docbook/gnc-locale-C.dtd
```diff
@@ -39,171 +53,555 @@
 
 <!-- guimenu/guisubmenu/guimenuitem to be used as names in text -->
 <!-- <!ENTITY % gmi.xxx "<guimenuitem><accel></accel></guimenuitem>"> -->
-<!-- <!ENTITY gmi. "%gmi.;"> -->
+<!-- <!ENTITY gmi.xxx.yy-zz "%gmi.xxx.yy-zz;"> -->
+<!-- Description: 
+     gm:  short for <guimenu></guimenu>, every entry in the menubar.
+     gmi: short for <guimenuitem></guimenuitem>, each menu item, also in submenus
+     gsm: short for <guisubmenu></guisubmenu>, describes the menu entry of submenu
```
fellen, line 60: How about `, the leaves of the menu tree`
```
    gm:  short for <guimenu></guimenu>, the entries in the menubar;
    gsm: short for <guisubmenu></guisubmenu>, the submenus;
    gmi: short for <guimenuitem></guimenuitem>, the leaves of the menu tree.

```
Perhaps one could add also `the labels of …`

fellen:
Danke!

Was mir gerade auffiel:
4.1. sollte `Das GnuCash Hauptfenster` heißen und eine kurze Intro bekommen wie sinngemäß
Meist werden Sie in diesem Fenster arbeiten. Einige Aufgaben werden aber spezialisierte Fenster verwenden.

Das Fenster besteht aus verschiedenen Bereichen…

PR #267, opened 2022-05-30
FO formatting, Added parameter to PdfTarget...

CWehli:
for improving the presentation of 'glosslist' elements.

Maybe there are better approaches, but the output is significantly improved. Compare also PR #246


fellen:
@CWehli can you check, if it is also desired for our other used list types?

CWehli:
@fellen, I believe that these are the only two types of lists (glosslist and variablelist) that require this special formatting.

CWehli:
In my opinion this PR can be closed, because with commit 5744acc the change was already entered, right?

fellen:
Right, already done as commit https://github.com/Gnucash/gnucash-docs/commit/5744accd3b93a4c594b06f21568bb01b69717557.

PR #269, opened 2022-06-08
Check images

gjanssens:
Extra checks for images. It will warn if a non-existent image is references in the documentation and notify if images in the figure directory are not referenced.

Both are potential issues:
* referencing non-existing images means the document isn't presented as intended
* having unreferenced images in the figures directory means we're needlessly installing files that are not unused, hence wasting disk space on users' computers

I started thinking of this issue after a [short IRC chat with @fellen](https://lists.gnucash.org/logs/2022/05/28.html#T04:25:28) recently. The discussion was about cmake not capturing when a new image was added. That problem can't be solved unless we explicitly list all used images in CMakeLists.txt files rather than the file glob we use currently.

This new check doesn't solve this issue either (it just can't), but it can be a useful tool to keep our used images in sync with our existing images. If we'd change our policy to explicitly list all images this check could be altered/extended to match existing images against the image list in CMakeLists.txt and against the image references in the documents themselves.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # CMakeLists.txt for GnuCash Documenation
 
-cmake_minimum_required (VERSION 3.5)
+cmake_minimum_required (VERSION 3.10)
 
 # This sets a number of environment variables we can use later on
 # The names of these variables start with PROJECT_ and gnucash-docs_VERSION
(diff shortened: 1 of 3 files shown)


fellen (approved):
I didn't look in all details, but the result is impressive—several forgotten images.

PR #270, opened 2022-06-12
dtd-docbook: Separation of the declaration of entities of the GUI

CWehli:
... from their translatable part. This is for the sake of clarity and
makes the translator's work easier.

Each gnc-locale-* file contains the translation of GUI Elements, the
gnc-gui-struct file defines the entities for menuitems and menuchoice.


fellen:
Thanks!

PR #271, opened 2022-06-16
[Bug 798432] manual ch_Transactions: Date field options

CWehli:


CWehli:
These and many more abbreviations are already described at the guide: https://www.gnucash.org/docs/v4/C/gnucash-guide/chapter_txns.html#txns-shortcuts1

PR #281, opened 2022-07-24
manual: Changed ref 'invest-stockprice-...' into 'tool-price-...'

CWehli:
According to [https://wiki.gnucash.org/wiki/Documentation_Improvement#Conventions](https://wiki.gnucash.org/wiki/Documentation_Improvement#Conventions), section '**ID attributes**':
Many elements can contain an inside of the document unique id="**chapter**-more-specific-context" attribute.


fellen:
Danke!

PR #285, opened 2022-08-12
Update ghelp target to install docs according to the xdg help specifi…

gjanssens:
…cation

The spec can be found here: https://www.freedesktop.org/wiki/Specifications/help-spec/
"ghelp:" is hardly used still on linux. The new standard is well supported by yelp.

What this does is
* rename the primary documents from gnucash-guide.xml/gnucash-help.xml to index.docbook
* update book id's for manuals to "index" (required for the spec)
* reshuffle the target output directories from share/gnome/help/(docname)/lang to share/help/(lang)/(docname)

Should be merged at the same time as "ghelp2help" in the "gnucash" repository.

TODO

- [ ] Solve viewing gnucash-help on the website. gnucash-help before 5.x references "help.html" as index page, while 5.x and later will use "index.html".


gjanssens:
My proposal to cope with this on the webserver is in the gnucash-htdocs beta branch. It will need some intervention from linas to unblock that branch on the webserver before we can test it though.

What it does it will change the viewdoc.phtml php code to check the requested doc version before it checks for the requested document itself. In between it will adjust to initial page to load for the help manual based on the version of the doc requested. If the version is "4" or older, "help.html" will be served, newer will server "index.html".

Note that as long as gnucash 5 is not officially released there will be no visible change as the help manual will only change in gnucash 5.

Note also that when gnucash 5 will be released, the "current", "maint" and "nightly" editions of docs will change from version "4" to version "5" and hence should be removed from the list of versions that will serve "help.html". I have added a comment in the code as a reminder as well.

jralls:
Shouldn't this rename AddGhelpTarget.cmake and the targets it creates to something else, maybe AddXDGHelpTarget.cmake etc.?

gjanssens:
Fair point.

gjanssens:
Do you want to keep the old AddGHelpTarget.cmake around or is one help target for linux sufficient ? (I vote for the latter...)

jralls:
> I vote for the latter...

Me too. @fellen ?

fellen:
I would prefer to 
1. rename any *help* to *manual*,
2. reorder the source directory structure according the XDG standard.

fellen:
> "ghelp:" is hardly used still on linux. The new standard is well supported by yelp.

ghelp was deprecated a decade ago. AFAIK we are the last project still using it.

cmake/AddChmTarget.cmake
```diff
@@ -20,7 +20,7 @@ function (add_chm_target docname lang entities figures)
            "-Dentities=\"${entities}\""
            -D HHC=${HHC}
            -P ${CMAKE_SOURCE_DIR}/cmake/MakeChm.cmake
-        DEPENDS ${entities} "${docname}.xml" "${CMAKE_SOURCE_DIR}/docbook/gnc-docbookx.dtd" ${figures}
+        DEPENDS ${entities} "index.docbook" "${CMAKE_SOURCE_DIR}/docbook/gnc-docbookx.dtd" ${figures}
```
fellen, line 23: As we finally have to use "index.docbook" IMHO we should rename in our sources *.xml to *.docbook.

jralls:
> rename any *help* to *manual*,

That makes no sense. The spec requires that the docs be installed in a directory named `help` and the spec is named the "Help Specification". 

> reorder the source directory structure according the XDG standard.

The source directory structure should be optimized for authors. It's the build system's job to move things around to the right place for whatever the target needs.

fellen:
> > rename any _help_ to _manual_,
> 
> That makes no sense. The spec requires that the docs be installed in a directory named `help` and the spec is named the "Help Specification".

There is one directory help with 2 docs: guide and manual.
When our original contributors have split off the guide, they should have renamed the rest to manual, too.
Any further use of "Help" is ambigous and since years confusing readers and contributors.

fellen:
> > reorder the source directory structure according the XDG standard.
> 
> The source directory structure should be optimized for authors. It's the build system's job to move things around to the right place for whatever the target needs.

1. For authors and translators it is easier to have everything below their language code:
 - data
 - docbook? the language specific parts, currently only GUI related
 - guide
 - manual
2. Cross-linking is easy done.
3. For maintainers it will be ways easier.

jralls:
> Source dir
>...
>* data
>* docbook?
>* guide
>* manual

OK, I don't have any problem  with that.  I guess that the `data` and `docbook?` files are ones that are duplicated in the current e.g. guide/de and manual/de directories. 

The alternative is #120 which uses gettext to do the translations instead.

jralls:
> There is one directory help with 2 docs: guide and manual.

You and @CWehli  already did that last summer in #276.

What has that to do with renaming `AddGHelpTargets.cmake`?

fellen:
> I guess that the `data` and `docbook?` files are ones that are duplicated in the current e.g. guide/de and manual/de directories.

about [data](https://github.com/Gnucash/gnucash-docs/blob/maint/data/README)
about [docbook](https://github.com/Gnucash/gnucash-docs/blob/maint/docbook/README)
We still did not test all possible solutions, e.g. "Can we use DTD catalogs and would they have benefits?"
Currently it could keep gnc-gui-${LANG}.dtd. Currently that files have one non-GUI entity at the end. If we get  more they should get their own file.

 I could think about a script merging
1. https://github.com/Gnucash/gnucash/tree/maint/po/${LANG}.po 
2. https://github.com/Gnucash/gnucash-on-windows/blob/master/inno_setup/gnucash-mingw64.iss

into gnucash-docs/docbook/gnc-gui-C.dtd to generate most of gnc-gui-${LANG}.dtd, but it is not trivial because `<accel>F</accel>ile` had to search msgid "_File"
de.po would return msgstr "_Datei" and then it had to insert `<accel>D</accel>atei`

> The alternative is #120 which uses gettext to do the translations instead.

That would block the current workflow: writing in German and later translating to English.

fellen:
> > I vote for the latter...
> 
> Me too. @fellen ?

I would suggest to rename it to AddDbkTarget.cmake
dbk is the short file extension for dokbook, which again is the target format.

jralls:
>I would suggest to rename it to AddDbkTarget.cmake
dbk is the short file extension for dokbook, which again is the target format.

Not exactly. It's the format of the files placed in the XDG help directories, but it's also the format of the source files so it's not terribly informative about what the target does.

jralls:
> I could think about a script merging

OK, but that's not really germane to this PR. By all means go write that script and make it into a separate PR.

jralls:
> That would block the current workflow: writing in German and later translating to English.

Only @CWehli, and not really block. He'd write in German, translate to English, then convert his German msgids into msgstrs. One extra step. Being able to expose the docs to weblate would pay that extra effort back 100 times over by having many more translations and having them stay more or less up to date.

fellen:
> > There is one directory help with 2 docs: guide and manual.
> 
> You and @CWehli already did that last summer in #276.
> 
> What has that to do with renaming `AddGHelpTargets.cmake`?

We have still targets of the form `{LANG}-gnucash-help-{TARGET}` — I got feedback terrible complicated.
I would prefer `{COMPONENT}-{LANG}-{TARGET}` with
`COMPONENT={manual|guide}`

fellen:
> > That would block the current workflow: writing in German and later translating to English.
> 
> Only @CWehli, and not really block. He'd write in German, translate to English, then convert his German msgids into msgstrs. One extra step. Being able to expose the docs to weblate would pay that extra effort back 100 times over by having many more translations and having them stay more or less up to date.

In the current state using a diff-editor for translation offers much more options:
bidirectional, merge new structures from de into C, …

jralls:
> We have still targets of the form {LANG}-gnucash-help-{TARGET} — I got feedback terrible complicated.
I would prefer {COMPONENT}-{LANG}-{TARGET} with
COMPONENT={manual|guide}

Not exactly. The target under discussion is
```
 add_custom_target("${lang}-${docname}-ghelp"
 ```

That's the only one that's germane to this PR, and the question on the table is about the `ghelp` part, the `TARGET` in your scheme. The `COMPONENT`/`docname` still being `help` instead of `manual` is your fault: You and @CWehli should have changed that when you renamed everything else.

Aside from that, your preference makes little sense: The *target* directory structure is `lang/docname` and having the build target reflect that is logical. It will be even more logical after merging your coming PR to align the source structure with the target structure. You are working on that, right?

fellen:
> > We have still targets of the form {LANG}-gnucash-help-{TARGET} — I got feedback terrible complicated.
> > I would prefer {COMPONENT}-{LANG}-{TARGET} with
> > COMPONENT={manual|guide}

Right, that was the old structure. So {LANG}-{COMPONENT}-{TARGET} is OK for me.
> Not exactly. The target under discussion is
> 
> ```
>  add_custom_target("${lang}-${docname}-ghelp"
> ```

Correct, but `ghelp` is a terrible name. BTW IIRC it stands for Gnome Help.
And `help` is not possible as it is an internal target listing all targets.

> That's the only one that's germane to this PR, and the question on the table is about the `ghelp` part, the `TARGET` in your scheme. The `COMPONENT`/`docname` still being `help` instead of `manual` is your fault: You and @CWehli should have changed that when you renamed everything else.

We didn't touch the whatever-make-magic. We dropped only a redundant part of the pathes gnucash-help/…/help_foo…

> Aside from that, your preference makes little sense: The _target_ directory structure is `lang/docname` and having the build target reflect that is logical. It will be even more logical after merging your coming PR to align the source structure with the target structure. You are working on that, right?

jralls:
>Correct, but ghelp is a terrible name. BTW IIRC it stands for Gnome Help.
And help is not possible as it is an internal target listing all targets.

Yes, it stands for Gnome Help. It's correct until this PR because that's what the target installs for: Gnome Help.

`${lang}-${docname}-help` should be OK, but you're right that we can't use `help` for the top-level target and it would be better for all of the target names to be consistent.

So we're back to the original question. I suggested `XDGHelp` because the target will install the files according to the XDG Help Spec. Is that OK with you?

fellen:
> So we're back to the original question. I suggested `XDGHelp` because the target will install the files according to the XDG Help Spec. Is that OK with you?

From my POV using also here the target format (DBK) would be easier to remember for new contributors. But it is not worth endless discussions. So XDGHelp is acceptable.

gjanssens:
Well, the two of you have been having a lively conversation while I was away :wink: 

Let's do the essential question first: what name to use for the new target. There is something to say for either:

- XDG is unambiguous wrt to the source files, but is a term that doesn't mean much for most contributors.
- Dbk is the exact opposite - the target output format is clearly described in the target name, however it may lead to confusion with the source format.

As you both more or less settled on XDG, I'll just go with that. If more discussion is needed, we can do so at length (eg on the devel mailing list) until we start rolling out 5.x pre-releases.

gjanssens:
>     1. rename any `*help*` to `*manual*`,
> 
Not for this PR, but I may make another one for that. However I'd need to know how far to go:

* source directory has already been renamed
* do you want the actual output to be renamed to gnucash-manual (directory for xdg spec and html output), gnucash-manual.pdf and so on
* do you want the build targets to be renamed from {LANG}-{docname}-target to {LANG}-{manual|guide}-target ? (so replace 'gnucash-guide' with 'guide' and 'gnucash-help' with 'manual' in all targets) ? Or stick with {docname} but use gnucash-manual as docname as suggested in my second bullet ?
* where else is gnucash-help used and disturbing ?

>     2. reorder the source directory structure according the XDG standard.

Depending on possible time opportunities I may look at this as well at some point. A major release as a good point in time to make such structural changes. I would suggest to merge such a PR only at the very last moment though as it would become a source of merge conflicts.

gjanssens:
> > > That would block the current workflow: writing in German and later translating to English.
> > 
> > 
> > Only @CWehli, and not really block. He'd write in German, translate to English, then convert his German msgids into msgstrs. One extra step. Being able to expose the docs to weblate would pay that extra effort back 100 times over by having many more translations and having them stay more or less up to date.
> 
> In the current state using a diff-editor for translation offers much more options: bidirectional, merge new structures from de into C, …

I agree with @jralls a po based workflow could be interesting and not necessarily making German-first editing more cumbersome. Granted we may need to adjust our tooling and way of thinking about writing documentation at that point for it to become equally efficient. But for the things you mention I believe a few small scripts (or cmake rules) might be all that's needed as a good replacement. I'm not pushing for this change currently though as it's highly unlike I'll have enough availability to carry _that_ to a successful transition myself.

fellen:
> > ```
> > 1. rename any `*help*` to `*manual*`,
> > ```
> 
> Not for this PR, but I may make another one for that. However I'd need to know how far to go:
> 
>     * source directory has already been renamed
> 
>     * do you want the actual output to be renamed to gnucash-manual (directory for xdg spec and html output), gnucash-manual.pdf and so on

Yes.
>     * do you want the build targets to be renamed from {LANG}-{docname}-target to {LANG}-{manual|guide}-target ? (so replace 'gnucash-guide' with 'guide' and 'gnucash-help' with 'manual' in all targets) ? Or stick with {docname} but use gnucash-manual as docname as suggested in my second bullet ?

The question is: While  the `gnucash` prefix is required on installation, does it make sense in build or can we just use the component(=directory names)? While developing —ninja this; ninja that …—the full name is primary noise.

>     * where else is gnucash-help used and disturbing ?

I hope, we catched all, @CWehli ?
 
> > ```
> > 2. reorder the source directory structure according the XDG standard.
> > ```
> 
> Depending on possible time opportunities I may look at this as well at some point. A major release as a good point in time to make such structural changes. I would suggest to merge such a PR only at the very last moment though as it would become a source of merge conflicts.

Shouldn't it simplfy the scripts?

fellen:
> > > > That would block the current workflow: writing in German and later translating to English.
> > > 
> > > 
> > > Only @CWehli, and not really block. He'd write in German, translate to English, then convert his German msgids into msgstrs. One extra step. Being able to expose the docs to weblate would pay that extra effort back 100 times over by having many more translations and having them stay more or less up to date.
> > 
> > 
> > In the current state using a diff-editor for translation offers much more options: bidirectional, merge new structures from de into C, …
> 
> I agree with @jralls a po based workflow could be interesting and not necessarily making German-first editing more cumbersome. Granted we may need to adjust our tooling and way of thinking about writing documentation at that point for it to become equally efficient. But for the things you mention I believe a few small scripts (or cmake rules) might be all that's needed as a good replacement. I'm not pushing for this change currently though as it's highly unlike I'll have enough availability to carry _that_ to a successful transition myself.

As long as we do many structural changes—convert `<ItemizedList>` into `<GlossList>`,`<OrderedList>` into `<Procedure>`… or moving often used terms into enities or snippet files MsgIds are not stable.

gjanssens:
> > ```
> > * do you want the actual output to be renamed to gnucash-manual (directory for xdg spec and html output), gnucash-manual.pdf and so on
> > ```
> 
> Yes.

Ok. Will need synchronized changes in gnucash-htdocs and gnucash repos though.
> 
> > ```
> > * do you want the build targets to be renamed from {LANG}-{docname}-target to {LANG}-{manual|guide}-target ? (so replace 'gnucash-guide' with 'guide' and 'gnucash-help' with 'manual' in all targets) ? Or stick with {docname} but use gnucash-manual as docname as suggested in my second bullet ?
> > ```
> 
> The question is: While the `gnucash` prefix is required on installation, does it make sense in build or can we just use the component(=directory names)? While developing —ninja this; ninja that …—the full name is primary noise.
> 
I'm fine with dropping them. I agree the target names are cumbersomely long and the gnucash bit in there doesn't make add distinguishing detail.

> > > ```
> > > 2. reorder the source directory structure according the XDG standard.
> > > ```
> > 
> > 
> > Depending on possible time opportunities I may look at this as well at some point. A major release as a good point in time to make such structural changes. I would suggest to merge such a PR only at the very last moment though as it would become a source of merge conflicts.
> 
> Shouldn't it simplfy the scripts?

Not really no. In order to only copy the files listed in CMakeLists.txt, we will have to keep most custom_commands and targets we have today. The only thing that would change is the order of directories in target destination specifications.

PR #286, opened 2022-08-15
Cmake track figures

gjanssens:
Changes our build scripts to hold explicit lists of figures per document. This should resolve the issue of cmake not rerunning when figures are added or removed.

It also extends the image checks to verify referenced images in documents are aligned with images listed in CMakeLists.txt.


gjanssens:
Maybe of note: the checks run on images have different error severities.

* When there are more images on the file system than referenced in the document, a status message is emitted
* When images are listed in CMakeLists.txt that are not referenced in the document, a status message is emitted
* When some images are referenced that are not present on the file system, a warning is emitted
* When some images are referenced that are not listed in CMakeLists.txt, a warning is emitted

I am tempted to flag the latter two as errors instead. That would however cause distcheck to fail currently as a few documents currently emit these warnings (guide-it, manual-it, guide-pt). Perhaps we should first get the whole of gnucash-docs to a point where these warnings are fixed and then mandate reference correctnes (that is make the warnings errors). Thoughts ?

fellen:
Thanks!

jralls:
We don't run distcheck often enough for it to be useful here, we need the missing files to show up in CI. It would be nice if files missing from dist were also flagged up in CI, but that's a separate issue. I think the last two should be errors and we should consider making the second one an error too. I don't think a count of files on the filesystem is terribly useful because those extra files are likely to be file editor artifacts, extra screenshots that didn't go quite right, or other local stuff that aren't of any real concern. A check the repository for files that aren't mentioned in the build system would be useful but also difficult to implement.

fellen:
At least I run distcheck before merging structural and other major changes.

The lists of <q>superfluous</q> files are a good help for maintenance work. As others could be annoyed one could consider to display them only if a verbose flag is set.

Some files like e.g. *.xfc are GIMP source files to create callout screens—images with annotations—as .png and not required in the build.

PR #287, opened 2022-08-19
Calculate price if importing securities

flywire:
https://bugs.gnucash.org/show_bug.cgi?id=797450#c9

### DIFF
--- a/manual/C/ch_Transactions.xml
+++ b/manual/C/ch_Transactions.xml
@@ -1932,7 +1932,8 @@ Translators:
 
                   <listitem>
                     <para><emphasis role="bold"><guilabel>Price</guilabel></emphasis>: The price for commodity transactions
-                      and/or transactions involving currency exchange;
+                      and/or transactions involving currency exchange. Price is not calculated by import and must be calculated
+                      using integers in the fraction (e.g. not 9912.53/89 use 991253/8900) and imported as a field.;
                     </para>
                   </listitem>
 


jralls:
I think this needs some more explanation in the introduction, perhaps as a note about importing transactions in more than one commodity. Although it's not mentioned in the particular bug reports, the CSV importer can also mess up transactions with more than two commodities unless the Transaction Commodity column is populated because it may pick the wrong split's commodity for the transaction commodity.  The Deposit and Withdrawal columns need to explicitly say that the amounts are treated as being in the transfer account's commodity and the conversion to the transaction currency is calculated by price * (deposit - withdrawal).

You might also mention that if a price is needed and the price column is empty or 0 that the importer will look for a price in the pricedb and fall back on 1 as a last resort.

flywire:
I agree the commodity explaination is fairly poor but I don't consider my experience with them sufficient to offer an explanation. I don't see that precludes an explanation of how to enter a price.

fellen:
> I think this needs some more explanation in the introduction, perhaps as a note about importing transactions in more than one commodity. Although it's not mentioned in the particular bug reports, the CSV importer can also mess up transactions with more than two commodities unless the Transaction Commodity column is populated because it may pick the wrong split's commodity for the transaction commodity. The Deposit and Withdrawal columns need to explicitly say that the amounts are treated as being in the transfer account's commodity and the conversion to the transaction currency is calculated by price * (deposit - withdrawal).
> 
> You might also mention that if a price is needed and the price column is empty or 0 that the importer will look for a price in the pricedb and fall back on 1 as a last resort.

To avoid redundancies the basic details should be explained in the multi-currency chapter and then it should be linked from here. And this secion should only get the import specific part.

@flywire see https://wiki.gnucash.org/wiki/Documentation_Improvement for details and ust reiterate what John told here.

flywire:
> the CSV importer can also mess up transactions with more than two commodities

When I saw that I realised I'm not competent with this functionality. I suppose it is purchase securities with foreign currency, not something I'd likely use and best left for someone else to explain.

PR #290, opened 2022-09-30
Clarify warning

felagund:
Related to this discussion: https://bugs.gnucash.org/show_bug.cgi?id=798624 . I think the curent wording is unspecific and the proposed wording is clearer.

### DIFF
--- a/manual/C/ch_Finance-Quote.xml
+++ b/manual/C/ch_Finance-Quote.xml
@@ -464,7 +464,7 @@ BEGIN failed--compilation aborted at (eval 303) line 1.
       &app;-data file without starting the GUI. Thus an automatic, regular updating of the price
       database is possible.
       <warning>
-        <para>Please note that the data file should not be open in another instance of &app; at the same time!
+        <para>Please note that if the data file is opened in another instance of &app; at the same time, the script will fail!
         </para>
       </warning>
     </para>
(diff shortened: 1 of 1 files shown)


manual/C/ch_Finance-Quote.xml
```diff
@@ -464,7 +464,7 @@ BEGIN failed--compilation aborted at (eval 303) line 1.
       &app;-data file without starting the GUI. Thus an automatic, regular updating of the price
       database is possible.
       <warning>
-        <para>Please note that the data file should not be open in another instance of &app; at the same time!
+        <para>Please note that if the data file is opened in another instance of &app; at the same time, the script will fail!
```
jralls, line 467: It's a command, not a script, though it might be contained in a script. I think it would be clearer still to say something like "The command will fail if it cannot get the lock on the data file, for example if the the data file is open in another GnuCash instance or if the last session for the file crashed."

fellen: Can you reworden your commit to
`Bug 798624 - Document how to check if GnuCash is running when updating quotes`
to get a link?

felagund: Yes, here: https://github.com/Gnucash/gnucash-docs/pull/290/commits/73c55917417f958abf312bc73b6628b07b4f8a43

jralls:
[Merged](https://github.com/Gnucash/gnucash-docs/commit/9b200d81d0ede8558fe5c1432102589d12e2fc80) after rebasing away the first commit and changing the summary to indicate the bug report.

PR #292, opened 2022-10-30
Multi-split csv elaboration

flywire:
For discussion


fellen:
I think it is OK. I only improved the markup in commit e479538.

Thanks!

PR #309, opened 2023-03-27
 (3) Set attribute of *list spacing="compact" for FO/PDF.

jralls:
Recreate #237

Depends on #310 

This is requested by @CWehli in https://github.com/Gnucash/gnucash-docs/pull/211#issuecomment-923815297.
This PR is for FO/PDF only.

I'll create another PR for HTML(s) because it is a bit complex to do.
It becomes clear by using external CSS. So I have to set some parameters, create CSS file, and change cmake files.


CWehli:
@usamimikamen-sk : 
I would like to take over some of your work and continue it. Do you have any objections?
I would also mention your name in the commit message.

PR #255, opened 2022-03-13
ch_Account-Actions: Backport the completely German version

CWehli:
After ch_Account-Actions was translated into German and at the same time
further improved in large parts and supplemented by missing sections,
the innovations are now backported.

For an easier review, each sections has an own commit.

PR #314, opened 2023-03-27
 ch_Account-Actions: Backport the completely German version 

jralls:
Recreate #255


CWehli:
@jralls, @fellen  : please check. It's the backport of the German improvements.

jralls:
@CWehli, @fell, ISTM this belongs in the T&CG. Any material that's here should be added there and this chapter deleted.

CWehli:
@jralls, @fell,
> ISTM this belongs in the T&CG. Any material that's here should be added there and this chapter deleted.

In principle, I agree, as this section does not describe the user interface.
However, I also find that this section helps the reader of the manual by describing the account types, because hereby the relationships between the individual accounts are explained.
I'm not hung up on the section, so if you think it fits better in the guide, I'll throw it out here and refer to [Guide Chapter 3, GnuCash Accounts](https://code.gnucash.org/docs/C/gnucash-guide/accts-types1.html) in the other sections instead.

jralls:
@CWehli, redundancy is bad. There's an explanation of the account types in the guide and because it's a fundamental element of accounting that's the right place for it. The manual should describe the selector and point to the section in the guide for more information about the different account types.

CWehli:
@jralls, @fellen :
I've removed the explanation of account types. Please merge and then i will go on.
The next step will be the backport of section 'New Account Hierarchy Setup'

fellen:
I just detected this PR.
@jralls I see it different. At least I am missing the translation of https://www.gnucash.org/docs/v4/de/gnucash-help/acct-types.html
Reason. While there are 5 types known from  accounting 1X1, there are a bunch of GnuCash specific accounts. Why are they special and how relate they to the standard hierachy? That should be answered before creating charts of account or operating on them.
They are technical, not didactical, and GUI related as they fill several dropdown lists.
See it as a user friendly variant of https://code.gnucash.org/docs/MAINT/group__Account.html.

jralls:
@fellen, rubbish. Most of the Guide is Gnucash-specific. The various account types are all described in detail in various parts of it and I have no objection to the detailed table being added to the end of 2.1.1, ideally enhanced with pointers to subsequent chapters where each account type is described with more detail.

FWIW my Accounting 101 (and that was its actual title in the catalog) class covered a lot more than the 5 types covered in Guide section 2.1.1. It was naturally business-oriented so it didn't mention credit cards or mutual funds, but it did cover current, real, fixed, and intangible assets, short-term and long-term liabilities, reserve accounts, suspense accounts, and contra-accounts, and the differences between cash and accrual accounting.

fellen:
I still see it different.
The manual is still missing the description of the `Chart of Accounts` view. And inside of that the tables belong to the column description of `Account Type`. Then the tables should be referenced in `Create or Edit an Account` as field description of  `Account Type`— for Create only the first table, for Edit both.

And I see nothing didactic in this overview, so the guide would be the wrong part.

jralls:
> The manual is still missing the description of the Chart of Accounts view. 

That's because it's in [the Guide](https://www.gnucash.org/docs/v5/C/gnucash-guide/basics-running-gnucash.html#basics-main2).

> And inside of that the tables belong to the column description of Account Type.

Yup. But instead that table is  on here in the manual without any context.

> And I see nothing didactic in this overview, so the guide would be the wrong part.

I dunno what you mean by "didactic"; in English it means [designed or intended to teach](https://www.merriam-webster.com/dictionary/didactic). That applies to both documents. If you're trying to say that the Guide should be only about the basics of accounting then that's just the [first section of the first chapter](https://www.gnucash.org/docs/v5/C/gnucash-guide/basics-accounting1.html). The rest of it is about how to use GnuCash.

fellen:
> The rest of it is about how to use GnuCash.

That should develope into the direction `Approachs to get the best/efficient results from using Gnucash` instead of repeating basic views.

jralls:
> That should develope into the direction Approachs to get the best/efficient results from using Gnucash instead of repeating basic views.

OK. If you're going to do a ground-up rewrite (not a bad idea), consider creating a single document from both pieces. The manual is conceptually somewhat obsolete: I think it was intended as a home for the "context-sensitive" documentation reached by the help buttons (hence the old name), but the industry has moved away from that to things like tooltips and that's how most new features are documented in GnuCash too. There's not a lot of point to reciting the menus and list-box contents anyway: Users can see that for themselves by using the program and it makes for very dry reading.

jralls:
Merged after a rebase. Thanks!

PR #315, opened 2023-04-11
some stock entry clarifications

getaaron:
I'm a new gnucash user and found a few things unintuitive. Here are some proposed doc improvements I think would've helped me:

- be a bit more descriptive when mentioning where the `NASDAQ` field came from
- link to the security editor manual page so the reader doesn't have to hunt for it
- replace the bit about how gnucash "has the following built in:" (since these categories are no longer automatically included) with 2 suggested examples
- explain why the stock transaction assistant menu item might be disabled

It's my first PR here, all feedback welcome.

### DIFF
--- a/C/guide/ch_invest.xml
+++ b/C/guide/ch_invest.xml
@@ -764,10 +764,9 @@ Income
                 </listitem>
 
                 <listitem>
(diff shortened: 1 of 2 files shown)


jralls:
Thanks for submitting a PR and sorry for ignoring it for 3 years. Unfortunately it was obsolete when you made it because we removed NASDAQ and all of the other US exchange default namespaces in release 4.10 in September 2022.

PR #319, opened 2023-05-03
Choosing a Register Style

flywire:
I'm offering this for discussion.

1. Double-line format is not clearly explained and easily misunderstood as having something to do with splits
1. Auto-Split Ledger style is more easily explained after Basic Ledger and Transaction Journal style, and the two-point list is removed since there are three options

I'm envisaging a discussion about Description/Notes/Memo in relation to Tags.

### DIFF
--- a/C/guide/ch_txns.xml
+++ b/C/guide/ch_txns.xml
@@ -82,42 +82,27 @@
       <title>Choosing a Register Style</title>
 
       <para>&app; offers several options for viewing your registers. The default style is <guilabel>Basic
-        Ledger</guilabel> mode, which displays only the summary of splits affecting the current
+        Ledger</guilabel>, which displays only the total of splits affecting the current
         account. This is the style that most closely resembles other popular personal financial
         packages. You can choose a different register style from the <menuchoice>
           <guimenu>View</guimenu>
-        </menuchoice> menu. There are two other view modes:
+        </menuchoice> menu.
       </para>
 
-      <itemizedlist>
-        <listitem>
-          <para><menuchoice>
-              <guimenu>View</guimenu><guimenuitem>Auto-Split Ledger</guimenuitem>
-            </menuchoice> style expands the current transaction automatically. As you highlight a new transaction in the
-            register, the transaction automatically expands to show all splits.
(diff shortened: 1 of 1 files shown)


C/guide/ch_txns.xml
```diff
@@ -82,42 +82,27 @@
       <title>Choosing a Register Style</title>
 
       <para>&app; offers several options for viewing your registers. The default style is <guilabel>Basic
-        Ledger</guilabel> mode, which displays only the summary of splits affecting the current
+        Ledger</guilabel>, which displays only the total of splits affecting the current
```
dawansv, line 85: It's only a total if we are dealing with a 1-to-many split, even there it's the 1 you are seeing. Or this could be a many-to-many split, in which case it's certainly not a total, it's just a split amount. So I think it's better to simply say: 
`which displays only the amount of the split affecting the current ledger.`
I know there is more to it with the transfer account being shown (or "split transaction" if there are more than 2 splits) but that's just more info than needed here I think.

dawansv: Also maybe we should say register instead of ledger. It seems the 2 are used but I see register used below as well.

dawansv:
> 1. Double-line

The use of "double" here certainly is not great considering the use of the same word in "double entry accounting" while it has nothing to do with it here. But I am not even sure what I would replace it with.

flywire:
**2.9.2.2. Choosing a Register Style**

GnuCash offers three register **View** styles, selectable from the View menu, each suited to different needs for examining register transactions.

* **Basic Ledger** (default): Displays only the summary of splits affecting the current account, resembling popular personal finance software for a simplified overview.
* **Auto-Split Ledger**: Automatically expands the current transaction as you highlight it, revealing all splits for detailed inspection without manual action.
* **Transaction Journal**: Shows all splits for every transaction in the register, mimicking a full accounting journal for comprehensive analysis.

**Double-Line Mode**: All styles support single-line or double-line view format. Select **View → Double Line** to expand each transaction to two lines, also displaying the transaction-level Notes field.

---

Text needs coding.

PR #321, opened 2023-06-13
docbook: Change all occurrences of the entity 'maint' to 'stable

CWehli:


fellen:
grep still delivers a few. Most are Update Notes, but one link exists:
@jralls I can no longer find the target of https://github.com/Gnucash/gnucash-docs/blob/cf5c288afcb2fdb8896a0f86ac3a4460beb8f78f/C/manual/ch_Transactions.xml#L2592 
@CWehli can you in a separate PR check, if the colors in that chapter are the same as in https://github.com/Gnucash/gnucash/blob/stable/gnucash/gnucash-fallback.css?
[Edit:[ or ask the author, @DaveC49 IIRC

jralls:
@fellen It was removed in https://github.com/Gnucash/gnucash/commit/065002190adef7734827a09a6551c5fd5d1095d5. That was on master, so it was still there when @goodvibes2 wrote https://github.com/Gnucash/gnucash-docs/commit/5251a91f239ebcd4621231079f7177cb427dd2e9 that added the note.

You could link https://github.com/Gnucash/gnucash/blob/stable/gnucash/gnucash-fallback.css or you could leave it off and say that colors might depend on theming and the user's gtk-3.0.css.

fellen:
> You could link https://github.com/Gnucash/gnucash/blob/stable/gnucash/gnucash-fallback.css

commit b4622a9

PR #322, opened 2023-07-09
The Zh (Simplified Chinese) Translation for the Part.1 of the Guide.

xuxinhang:
This PR includes:
1. the text translation for the Guide (`/zh/guide`) based on commit `3baaf9`
2. updated CMake files to apply customized XSLT files to build production, which is necessary to meet the Chinese typesetting specification and availability of CHM searching function, and so on. The inner CMake function signatures is changed compatibly.

The build production looks fine.

And some questions:
1. Should `add_subdirectory(zh)` be enabled?
2. Which branch should this ongoing work be merged into? `maint` seems gone.
3. Or anything I should supplement.

Maybe better to discuss these questions above via maillist but there seems no such one.

Thx.

### DIFF
--- a/C/guide/index.docbook
+++ b/C/guide/index.docbook
@@ -81,6 +81,14 @@
 
     <revhistory>
       <revision>
+        <revnumber>&appname; Tutorial and Concepts Guide 5.2</revnumber>
+        <date>25 June 2023</date>
+        <revdescription>
+          <para role="author">Multiple authors</para>
+          <para role="publisher">&appname; Documentation Team</para>
+        </revdescription>
+      </revision>
+      <revision>
         <revnumber>&appname; Tutorial and Concepts Guide 5.1</revnumber>
         <date>30 April 2023</date>
         <revdescription>
--- a/C/manual/index.docbook
+++ b/C/manual/index.docbook
@@ -44,8 +44,16 @@
 
     <revhistory>
       <revision>
+        <revnumber>&appname; Help Manual 5.2</revnumber>
+        <date>25 June 2023</date>
+        <revdescription>
+          <para role="author">Multiple authors</para>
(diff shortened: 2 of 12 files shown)


jralls:
@xuxinhang A great start. I'll look over the CMake changes later, but to your questions:

>    Should add_subdirectory(zh) be enabled?

Yes, so that the build is tested in CI. I think you should get a couple more chapters done before we add it to the distribution, but that's a separate change that you don't need to worry about.

>    Which branch should this ongoing work be merged into? maint seems gone.

We're currently using only one branch, `stable`.  

>    Or anything I should supplement.

Yeah, back up a step. The PR should be that you want to merge `translation-zh` into `stable`. I think that means you need to close this and make a PR; the target (i.e. `stable`) branch can be changed but AFAIK the source branch can't. Don't do any merges, that's our job. It doesn't look like there's any need to rebase your cmake changes, nothing except the version number has changed from the commit on which you based your `translation-zh` branch.

Be sure to get `ch_import.xml`, `ch_Tools_Assistants.xml`,  and `ch_Transactions.xml` from stable before you translate them so that you're in sync with the latest English version.

xuxinhang:
Thanks for your reply.

I have removed the merge commit. I will open another PR.

As for the changes in English version, the current focus is the Guide, so the `C/manual/ch_Tools_Asssistants.xml` and `C/manual/ch_Transactions.xml` don't matter. And the translation work of `C/guide/ch_import.xml` would be scheduled later, due to it marked as "outdated". (Maybe somebody enthusiastic would rewrite or improve this chapter in the next few months, I guess.)

CMake build flow supporting customized xslt is the key to produce the easy-to-read Chinese publishment. Further discussion could proceed, maybe in the new PR.

PR #323, opened 2023-07-14
The Zh (Simplified Chinese) translation for the Guide and language-specific XSLT supporting.

xuxinhang:
This PR is re-created according to #322, which includes:

1. the text translation for Part 1 of the Guide (/zh/guide) synced with the latest English version.
2. updated CMake flow to apply customized XSLT files to build production, which is necessary to meet the Chinese typesetting specification and availability of CHM searching function, and so on. The inner CMake function signatures is changed compatibly.(Or in any better way?)
3. some improvement on English version.


C/guide/ch_txns.xml
```diff
@@ -1651,11 +1651,11 @@
       <para>But only having data available on the computer screen will not make your accountant happy, therefore
         &app; comes with a large number of highly customizable reports.
       </para>
-<!-- Add this note when the reports chapter is ready
+
```
fellen: Right, the note should be enabled.
Perhaps we should convert the comment into a translator tip
`<!-- Translators: Add this note when the reports chapter is ready -->`
Feel free to improve the text.

C/guide/ch_basics.xml
```diff
@@ -1674,7 +1674,7 @@ Translators:
       </note>
 
       <tip>
-        <para>On Unix and &mac;, the <keycap>~</keycap> symbol means the
+        <para>On Unix and &mac;, the <filename>~</filename> symbol means the
```
fellen: What is wrong with `<keycap>~</keycap>`?
On keyboards for latin writing one or  more keys labeled <q>-</q> exist.

C/guide/ch_txns.xml
```diff
@@ -336,7 +336,7 @@
             account name and &app; will fill in the remaining part of the name. Typing the separator
             character at any time will complete the current level of the account name, leaving the
             cursor positioned to start the next level of account name. For example, typing
-            <keycap>A:C</keycap> might complete to the <emphasis>Assets:Checking</emphasis> account.
+            <guilabel>A:C</guilabel> might complete to the <emphasis>Assets:Checking</emphasis> account.
```
fellen: Right, formally correct would be` <keycap>A</keycap><keycap>:</keycap><keycap>C</keycap>` perhaps capsuled in `<keycombo action='seq'>…</keycombo>`.

Perhaps you want to convert `<emphasis>Assets:Checking</emphasis> account` into `<guilabel>Assets:Checking</guilabel> account`? But  then we should use it consistenly–—at least in the whole chapter.

fellen (changes requested):
This is my review of the english section. Can you move it into a separate PR?
Reason: other translators do not expect changes on C in the chinese translation.

xuxinhang:
Thanks. I have removed the changes on English version. I'll open another PR version someday.

C/guide/ch_basics.xml
xuxinhang: Maybe it would be better to see it as a symbol in the os file path, instead of a physical keycap on the keyboard.

C/guide/ch_txns.xml
fellen: Fixed in commit 24a91f0. Perhaps you want to rebase this PR?

xuxinhang: Commits rebased.

(But, by the way, maybe you have known, it's unreasonable to use `<keycombo>` to instruct Chinese reader how to enter characters, because there is no correspondence between a single Chinese character and a single physical keycap. So I will express this meaning with `<emphasis>X:Y</emphasis>` in translated texts.)

Any other problems? (including the CMake flow)

fellen: > (But, by the way, maybe you have known, it's unreasonable to use `<keycombo>` to instruct Chinese reader how to enter characters, because there is no correspondence between a single Chinese character and a single physical keycap. So I will express this meaning with `<emphasis>X:Y</emphasis>` in translated texts.)

I know almost nothing about input methods for CJK scripts,  Perhaps you could create a wiki page —and link it in https://wiki.gnucash.org/wiki/Documentation_Improvement and other I18N/L10N pages—

1. explainig the differences compared with latin scripts
2. develope a rule set for the portation.  BTW: How is it solved in ja/? Tip: call `kdiff3 zh/guide/index.dokbook ja/guide/index.dokbook` to compare them.

jralls: @fellen, Japanese and Chinese input methods work similarly. The main difference is that Japanese has an alphabetic form called [Hiragana](https://en.wikipedia.org/wiki/Hiragana) (it actually has two but the other one, [Katakana](https://en.wikipedia.org/wiki/Katakana), is used for foreign words that don't generally map to Chinese characters, called Kanji in Japanese) so a Japanese using a Japanese keyboard will type in Hiragana or using a Roman keyboard in a literation called [Romaji](https://en.wikipedia.org/wiki/Romaji) while a Chinese will type in a Romanization, nearly always [Pinyin](https://en.wikipedia.org/wiki/Pinyin). The IME will use the entry to present a list of possible character completions in a drop-down list that changes as more letters are typed, somewhat like, though vastly more complex than, the typeahead search in the register.

Korean is written in an alpha-phonic system called [Hangul](https://en.wikipedia.org/wiki/Hangul) so the input method is a simple keyboard mapping. 

> call kdiff3 zh/guide/index.dokbook ja/guide/index.dokbook to compare them.

Since the Japanese translation is now almost 12 years old that's very unlikely to produce anything useful.

fellen:
The current build failures are caused by missing entites. They should be defined in docbook/ gnc-docbookx.dtd or gnc-gui-zh.dtd.
[Edit:] @CWehli is permanently improving gnc-docbookx.dtd and gnc-gui-C.dtd. The latter should get translated into gnc-gui-zh.dtd.

C/guide/ch_txns.xml
xuxinhang: Japanese is based on the Chinese character, but using phonetic symbols for publishment is acceptable. So, you can input Chinese character used in Japanese with phonetic symbols.

Chinese character used in Chinese can be input via phonetic symbols (Pinyin in Mainland, Zhuyin in Taiwan) or structure components (Wubi/Zhengma in Mainland, Cangjie in Taiwan), which both are mapped to US keyboard.

Modern Korean has dropped the Chinese character. The Korean character only shows exact pronunciation and is just like what you write down phonemes into a small square area. You can imagine a keyboard layout to input international phonemes.

Modern Vietnamese is another language that has dropped the Chinese character. It has its own alaphbet which also represents a pronunciation rule like Korean, but its writting is just like Latin - one letter by one letter from left to right.

xuxinhang:
Fixed and another two translated chapters.

I suppose my draft commit was occasionally pushed, with serveral semicolons missed. But it now seems fine, anyway.

The entities in `gnc-gui-C.dtd` such as `&gmi.xx.yy` seem rarely used in `C/guide` or `C/manual`. Is it a future plan to replace `<guilabel>` or `<guimenu>` with `&gmi.xx.yy`?

fellen:
> The entities in `gnc-gui-C.dtd` such as `&gmi.xx.yy` seem rarely used in `C/guide` or `C/manual`. Is it a future plan to replace `<guilabel>` or `<guimenu>` with `&gmi.xx.yy`?

@CWehli ?

CWehli:
> The entities in `gnc-gui-C.dtd` such as `&gmi.xx.yy` seem rarely used in `C/guide` or `C/manual`. Is it a future plan to replace `<guilabel>` or `<guimenu>` with `&gmi.xx.yy`?

Yes, in the long run the entities should be used in both documentation (guide and manual). I have already done that in the de/manual.
I still want to develop the entities for `<guilabel>`.

C/guide/ch_txns.xml
fellen: I compiled [GnuCash Wiki:CJK Input Methods](https://wiki.gnucash.org/wiki/CJK_Input_Methods) from your comments. Feel free to improve the page!

CMakeLists.txt
```diff
@@ -75,6 +75,12 @@ set(JAPANESE_MINCHO_TTF "ume-tmo3.ttf" CACHE STRING "Mincho TrueType font used f
 set(JAPANESE_GOTHIC_TTF "ume-tmo3.ttf" CACHE STRING "Gothic TrueType font used for Japanese pdf")
 set(japanese_fontdir "${CMAKE_SOURCE_DIR}/fonts/truetype" CACHE STRING "Directory to search for Japanese fonts")
 
+# Set the default XSL file path
+set(BASE_XSL_HTML "${CMAKE_SOURCE_DIR}/xsl/general-customization.xsl")
+set(BASE_XSL_PDF "${CMAKE_SOURCE_DIR}/xsl/general-fo-customization.xsl")
+set(BASE_XSL_EPUB "${CMAKE_SOURCE_DIR}/xsl/general-epub-customization.xsl")
+set(BASE_XSL_CHM "${CMAKE_SOURCE_DIR}/xsl/general-chm-customization.xsl")
+
```
fellen: I assume this change and most others in this directory are improvements of our build system. I would prefer to have them in a separate Pull Request. That would make it easier to review them by the build system experts like @gjanssens .
But as I was able to run `ninja distcheck` they seem not to break our system fatally. ;-)

CMakeLists.txt
```diff
@@ -182,6 +188,7 @@ add_subdirectory (it)
 add_subdirectory (ja)
 add_subdirectory (pt)
 #add_subdirectory (ru)
+add_subdirectory (zh)
```
fellen, line 191: This change belongs into this PR.

fellen (commented):
See my comments inline.
[Edit:]
Before translating the next chapter, you should consider to localize the images in your already translated text, see [Screenshots And Other Images](https://wiki.gnucash.org/wiki/Documentation_Improvement#Screenshots_And_Other_Images).

xuxinhang:
> I assume this change and most others in this directory are improvements of our build system. I would prefer to have them in a separate Pull Request.

Sure. But, I wonder how to create another PR. Does that mean to cherry-pick these commits into another branch? However, if so, I'm afraid I have to keep these changes in my working branch for text translation, otherwise it would be harder to checking the translated texts and the display effect of publishments. Could you share me some tricks for such case that one PR depends on another PR?

> you should consider to localize the images in your already translated text

Yes, I have been realizing the screenshot-translation undone. I have noticed there is some outdated translation in GnuCash's PO files. So I decide to combine screenshot translation with the improvement work of GUI translation and the final text checking, after finishing the text translation. Besides, the Gtk3 GUI on Windows seems somewhat incongruity, so maybe I would prepare Gnome for screenshots later.

fellen:
> Sure. But, I wonder how to create another PR. Does that mean to cherry-pick these commits into another branch? However, if so, I'm afraid I have to keep these changes in my working branch for text translation, otherwise it would be harder to checking the translated texts and the display effect of publishments. Could you share me some tricks for such case that one PR depends on another PR?
> 
> >

In theory:
You cherry-pick  the general changes in a separate branch and rebase this branch on the new one. The duplicate commits should the disappear.
Important: Edit your first comment with a note "requires PR #xxx" as reminder for us.

For now:
@gjanssens , @jralls can you review commit f92c579?

fellen:
> Yes, I have been realizing the screenshot-translation undone. I have noticed there is some outdated translation in GnuCash's PO files. So I decide to combine screenshot translation with the improvement work of GUI translation and the final text checking, after finishing the text translation. Besides, the Gtk3 GUI on Windows seems somewhat incongruity, so maybe I would prepare Gnome for screenshots later

OK, but zh/guide/figures/basics_AccountRelationships.svg is no screenshot and we could reuse it for https://www.gnucash.org/features.phtml?lang=zh_CN in https://github.com/Gnucash/gnucash-htdocs/tree/master/images/features

jralls:
@fellen, why isn't separate commits sufficient? Once its merged the PR identity is mostly lost. Plus those commits are a necessary part of adding a Chinese translation.

CMakeLists.txt
gjanssens: For me it's fine they are part of this PR as the build system changes are in their own commit.

(BTW don't flatter me too much :wink: )

cmake/AddGncDocTargets.cmake
```diff
@@ -22,21 +22,42 @@ function (add_gnc_doc_targets lang entities figures)
         add_dependencies(${lang}-check "${lang}-${targetbase}-check")
     endif()
 
+    set(xslargs "${ARGN}")
```
gjanssens: Your extra code does what you intended. It is however missing a final bit: the newly added xsl documents should somehow be added to the list of files to distribute. Otherwise the release tarball would fail to build (and you can probably verify this by running make/ninja dist-check, which I haven't ATM).

There are a few ways to accomplish this:
One would be to add the files in the `add_to_dist` call in `zh/CMakeLists.txt`. That's a simple and straightforward solution. It would mean however users would have to remember adding this for each new xsl file. I had written the build system such that anything in the entities or figures lists would automatically be added to the dist files. For consistency I would prefer to have a similar solution for xsl files.

Proposal:
Extend the function signature of `add_gnc_doc_targets` to also have an explicit `xsl_files` parameter and use that instead of `${ARGN}`. Your parsing code here would remain the same for the normal build scripts.
But for the `add_to_dist` further down in this CMakeLists.txt file, you reparse to extract all the xml files to add for distribution.

In the calling CMakeLists.txt files (like `zh/guide/CMakeLists.txt`), formally set a variable called `xsl_files` and pass that to the call to `add_gnc_doc_targets`. So you'd get

```
set(xsl_files
     "xsl:pdf" "${CMAKE_CURRENT_SOURCE_DIR}/../xsl/zh-fo.xsl"
     "xsl:chm" "${CMAKE_CURRENT_SOURCE_DIR}/../xsl/zh-chm.xsl")

add_gnc_doc_targets(${lang} "${entities}" "${figures}" "${xsl_files}")
```

While that's functionally equivalent to what you did, it makes it more clear the function accepts a list of xsl files (with type spec) as extra parameter. That would make it a good example other languages could follow in the future.

Note: you have added several new files, but most of the others will be picked up by the build system code already, albeit after an explicit rerun of cmake (and perhaps even one that's not using the cached variables). From what I can see from direct review, only the two xsl files were missing.

fellen:
> why isn't separate commits sufficient?

One benefit: Geert would have seen an <q>Improve build system</q> PR before I ringed the bell.

cmake/AddGncDocTargets.cmake
fellen: @xuxinhang Do you need some help for this?

xuxinhang: Thanks. Fine. I'm just kinda overloaded this week. Sorry for my late reply.

It's a little confusing that whether the default xslt files in `/xsl` imported by `/zh/xsl/*.xsl` should be added to dist. I suppose it makes no sense to only add language-specific xslt files to dist due to breaking down without these base xslt files, but they are rather numerous.

Besides, the function signature has been modified according to your proposal.

xuxinhang:
> OK, but zh/guide/figures/basics_AccountRelationships.svg is no screenshot and we could reuse it for https://www.gnucash.org/features.phtml?lang=zh_CN in https://github.com/Gnucash/gnucash-htdocs/tree/master/images/features

Thanks for your reminding. I will translate these screenshots and diagrams used by website too. A good beginninh is now at https://github.com/Gnucash/gnucash-htdocs/pull/201 

> One benefit: Geert would have seen an "Improve build system" PR before I ringed the bell.

I will begin to re-sort commits after build system changes agreed. To cherry-pick too frequently seems somewhat dangerous.

cmake/AddGncDocTargets.cmake
gjanssens: The common xsl files are added to dist here:
https://github.com/Gnucash/gnucash-docs/blob/98a176d508852c9f0295c3fc19141bf6dd686a11/CMakeLists.txt#L186

The fact they are not explicitly listed could be considered a design flaw (partly by my own doing :sad: ). I think all files that are handwritten as part of the gnucash-docs should be individually listed before adding them to dist. This is different for the full stylesheet that was imported (and should ideally not be altered by us).

fellen: @gjanssens can you add a section about adding xsl files to https://wiki.gnucash.org/wiki/Documentation_Improvement#Add_Extra_Files?
OTOH I assume adding fonts and xslts is more a task for https://wiki.gnucash.org/wiki/Documentation_Translation?

fellen:
There is also translatable text in [zh/guide/index.docbook](https://github.com/xuxinhang/gnucash-docs/blob/a4f97475dbe4efe0c15cf559f0880f06bd207291/zh/guide/index.docbook#L555). In yelp it is shown on the `About` page.
You should also uncomment and update [Lines 60-65 there](https://github.com/xuxinhang/gnucash-docs/blob/a4f97475dbe4efe0c15cf559f0880f06bd207291/zh/guide/index.docbook#L60-L65).

I see more untranslated sections and will try to mark them inline.

zh/guide/ch_oview.xml
```diff
@@ -0,0 +1,525 @@
...
+        </listitem>
+
+        <listitem>
+          <para><emphasis>数据导入方法</emphasis>：
+          除了手动输入，&app;支持许多自动导入交易记录的方式。要是你可以在线访问银行账户，这一特性就特别实用，因为多数银行和信用卡公司都支持下述导入方式中的至少一种。如此一来，少一些时间录入数据，就可以多一些时间分析统计结果。
+          </para>
+
+          <itemizedlist>
+            <listitem>
+              <para><emphasis>Quicken Import File (QIF)</emphasis>:  Quicken <acronym>QIF</acronym> style files, a
```
fellen, line 78: Perhaps you want to read https://code.gnucash.org/docs/C/gnucash-guide/gnc-gloss.html, the original english output of your coming https://github.com/xuxinhang/gnucash-docs/blob/translation-zh/zh/guide/gnc-glossary.xml to understand this terms?

gjanssens (commented):
The build system changes look fine to me now.

Note that running make/ninja distcheck will point out an unused figure in the new translation:
- figures/txns_reconcile_window3.png

cmake/AddGncDocTargets.cmake
gjanssens: I don't know what you want me to add. With the changes @xuxinhang made to the build system, adding an xslt file is very similar to adding a figure. I don't immediately see that documented either and am tempted to just leave this at "learn by example".

fellen:
Thanks!

A few future tasks:

- [ ]  In between there have been a few [changes in the english guide](https://github.com/Gnucash/gnucash-docs/commits/stable/C/guide). I suspect they cause
```
zh/guide/ch_loans.xml:403: parser error : Entity 'url-gh-gc-maint' not defined
          <ulink url="&url-gh-gc-maint;&file-finc;"><filename>&file-finc;</filen
                                       ^
zh/guide/appendixa.xml:155: parser error : Entity 'url-src-maint' not defined
            <para>in our repository: <ulink url="&url-src-maint;contrib/xslt"/>

```
in `clear; ninja zh-guide-epub` and in `clear; ninja zh-guide-mobi 
`
```
Referenced file 'OEBPS/libgnucash/app-utils/calculation/fin.c' not found
Referenced file 'OEBPS/contrib/xslt' not found
```

Can you update the chinese version?
- [ ] running `clear; ninja zh-guide-pdf` i see a few
```
8月 20, 2023 2:04:46 下午 org.apache.fop.apps.FOUserAgent processEvent
WARNING: Font "Bitstream Vera Sans,normal,400" not found. Substituting with "any,normal,400".
8月 20, 2023 2:04:46 下午 org.apache.fop.apps.FOUserAgent processEvent
WARNING: Font "Source Han Sans CN,normal,400" not found. Substituting with "any,normal,400".
8月 20, 2023 2:04:46 下午 org.apache.fop.apps.FOUserAgent processEvent
WARNING: Glyph "收" (0x6536) not available in font "Times-Roman".
8月 20, 2023 2:04:46 下午 org.apache.fop.apps.FOUserAgent processEvent
WARNING: Glyph "入" (0x5165) not available in font "Times-Roman".
…
8月 20, 2023 2:04:50 下午 org.apache.fop.apps.FOUserAgent processEvent
WARNING: Glyph "−" (0x2212, minus) not available in font "Courier-Bold".
8月 20, 2023 2:04:51 下午 org.apache.fop.apps.FOUserAgent processEvent
INFO: Rendered page #126.
```

xuxinhang:
I'm back after another three busy weeks. I'm glad this PR merged and I'll contiune my translation work via another PR. 

------------

> There is also translatable text in zh/guide/index.docbook

Thanks for your reminding, I'll start to translate these texts soon. 

> figures/txns_reconcile_window3.png

Fixed. There are some untranslated texts.


> In between there have been a few changes in the english guide.

Thanks for your fixing `&url-gh-gc-maint;`. And other changes of C/guide have been synced to my working branch.

> in clear; ninja zh-guide-epub and in clear; ninja zh-guide-mobi  

I have been building all products (except for CHM) in Linux and all things seem fine. I try to search `fin.c` in terminal outputs but nothing. Could you share your building environment with me? Mingw32, Msys2 or anything else?

> running clear; ninja zh-guide-pdf i see a few

(Wow, are your using Windows with language as Chinese?)

There seems be some font missing in the log. The so-called missing "Source Han Sans" is due to incorrectly assigned font family names in SVG file, and I have fixed it. I'm still working out where the missing "Courier-Bold" comes from.

PR #327, opened 2023-08-31
728910: Added more documentation for Budget Balance Sheet

datajunkie-eng:
Here is a stab at some better documentation for the Budgeted Balance Sheet.  

Here's what it looks like:
bbs_docscreenshot

[Current Version of the doc](https://www.gnucash.org/docs/v4/C/gnucash-guide/budget_reporting1.html).


C/guide/ch_budgets.xml
```diff
@@ -204,10 +204,95 @@
     <sect2>
       <title>Budgeted Balance Sheet</title>
 
-      <para>The budgeted balance sheet is similar to the balance sheet. Both show the assets, liabilities, and
-        equity. The difference is that the balance sheet is based on historical data, and the
-        <emphasis>budgeted</emphasis> balance sheet is based on the predictions made in the budget.
+      <para>The Budgeted Balance Sheet is similar to the Balance Sheet (see <xref linkend="rpt_balancesheet"/>). Both show the assets, liabilities, and
+        equity. The difference is that the Balance Sheet is based on historical data, and the
+        <emphasis>budgeted</emphasis> balance sheet is based on the predictions made in the budget.  The report
```
fellen: As we are formalizing the docs —see [Text conventions](https://wiki.gnucash.org/wiki/Documentation_Improvement#Text_conventions)— can you mark the first part as abstract and the description as glosslist?
```
<abstract><para>The <guilabel>Budgeted Balance Sheet</guilabel> is […] 
predictions made in the budget.</para></abstract>
<para>The report  contains the following elements:
  <glosslist>
  <title>Elements of Budgeted Balance Sheet</title>
  <glossentry>
    <glossterm>>Assets Section</glossterm>
      <glossdef>
        <para>The value shown for each account […]
        </para>
      </glossdef>
    </glossentry>
    …
  <glosslist>
```
See [de/manual/ch_Tools_Assistants.xml](https://github.com/Gnucash/gnucash-docs/blob/stable/de/manual/ch_Tools_Assistants.xml) for advanced usage of glosslists.

C/guide/ch_budgets.xml
```diff
@@ -204,10 +204,95 @@
...
-        equity. The difference is that the balance sheet is based on historical data, and the
-        <emphasis>budgeted</emphasis> balance sheet is based on the predictions made in the budget.
+      <para>The Budgeted Balance Sheet is similar to the Balance Sheet (see <xref linkend="rpt_balancesheet"/>). Both show the assets, liabilities, and
+        equity. The difference is that the Balance Sheet is based on historical data, and the
+        <emphasis>budgeted</emphasis> balance sheet is based on the predictions made in the budget.  The report 
+        contains the following elements:
+
+        <itemizedlist>
+        <listitem>
+          <para><emphasis>Assets Section</emphasis> - The value shown for each account is its balance as of the start date of the selected budget, modified by the projected changes in this budget.  So if all goes according to the budget, these balances would be the same as if you were to run a standard Balance Sheet as of the end date of the budget period.
```
fellen: After you finished your edts can you please run [xmlformat](https://github.com/Gnucash/gnucash-docs/blob/stable/util/xmlformat/README) over the file?
BTW runoff style (2 spaces after a full stop) is obsolete, but get also fixed by xmlformat.

C/guide/ch_budgets.xml
```diff
@@ -204,10 +204,95 @@
...
+            <listitem>
+              <para><emphasis>Total Equity</emphasis> - Existing Equity + New Equity</para>
+            </listitem>
+          </itemizedlist>
+
+          </para>
+        </listitem>
+
+        <listitem>
+          <para><emphasis>Total Liabilities + Equity</emphasis> - Liabilities + Equity.  Equals Total Assets
```
fellen: Perhaps use <`equation`> also here?

C/guide/ch_budgets.xml
```diff
@@ -204,10 +204,95 @@
...
+            <listitem>
+              <para><emphasis>Allocated assets</emphasis> - the sum of all Asset row totals projected in the budget.</para>
+            </listitem>
+            <listitem>
+
+              <para><emphasis>Unallocated Assetss</emphasis> - This formula is:
+              <screen>
+          New Retained Earnings (all projected expenses subtracted from all projected income, per this budget) 
+        - Allocated Assets (above)
+        - New Liabilities</screen>
```
fellen: [Screens](https://tdg.docbook.org/tdg/4.5/screen-x) are processed as verbatim. You don't want the indentation. Start in column 1.

BTW did you consider to use [equation](https://tdg.docbook.org/tdg/4.5/equation-x) here and later?

fellen (commented):
A nice amendment, thanks!
A few notes:

- This is —like the previous report in that file— the technical description of the report. So they would belong into https://github.com/Gnucash/gnucash-docs/blob/cacf2e84336dea4e9df6f094e25ae8b5e30b7475/C/manual/ch_Reports.xml#L3775
Do you think,you can move them there and add here a link 
`[…] <ulink url="&url-docs-C;manual/report-classes.html#report-budget"><quote>Budget Reports</quote> in the manual</ulink>`?
Ideally that is done in 2 commits
  1. move the old report and add the link,
  2. add your new report.

- a few notes inline.

datajunkie-eng:
@fellen , I *think* I've addressed your feedback.  

I'm not really liking how the equations are formatting:
image

Maybe there's an easy way to make them look better?  I'll keep tinkering with it.

jralls:
> Equation 9.3

Use [`informalequation`](https://tdg.docbook.org/tdg/4.5/informalequation) to get rid of the `Equation XX` titles. Those are for math textbooks.

But better to describe it in words, e.g. "The sum of existing equity and new equity". And "Total Liabilities + Equity: Total Liabilities + Equity" is redundant. Use "self explanatory" instead.

datajunkie-eng:
@jralls , yep, I like this better.  Thanks!

image

C/manual/ch_Reports.xml
```diff
@@ -3778,7 +3778,179 @@ Income (type INCOME)
...
+                only goes up by $25 and you pay your credit card down by $26 (a combined increase of
+                only $51), you have a projected $49 of extra proceeds in the Income Statement that
+                doesn't have a corresponding increase in the Balance Sheet. If everything goes
+                exactly according to <emphasis>this</emphasis> budget, <emphasis>as-is</emphasis>,
+                either the Balance Sheet will grow by $49/month more than you're projecting (the
+                projected income/expenses are right), or the Income Statement for the whole period
+                will show a $51/month gain instead of $100 (the projected assets/liabilities are
+                right). This formula is:
+                <informalequation>
+                  <mathphrase>Unallocated Assets = New Retained Earnings - Allocated Assets - New
```
fellen: Nitpick from <q>Hyphens and dashes</q> in [Text conventions](https://wiki.gnucash.org/wiki/Documentation_Improvement#Text_conventions):
`New Retained Earnings &minus; Allocated Assets &minus; New`
There multiple other occurrences.

C/manual/ch_Reports.xml
```diff
@@ -3095,7 +3095,7 @@ Income (type INCOME)
                 </listitem>
 
                 <listitem>
-                  <para><guilabel>Report's currency</guilabel>: Select the currency to display the values in - see
+                  <para><guilabel>Report's currency</guilabel>: Select the currency to display the values in &minus; see
```
fellen: Caution, this should be `—` (`&mdash;` or enter [AltGr]+[Shift]+[-]).
In this case it is obvisious by the following <q>see …</q>.
[edit:] multiple occurrences

C/guide/ch_budgets.xml
```diff
@@ -206,7 +206,7 @@
 
       <para>The budgeted balance sheet is similar to the balance sheet. Both show the assets, liabilities, and
         equity. The difference is that the balance sheet is based on historical data, and the
-        <emphasis>budgeted</emphasis> balance sheet is based on the predictions made in the budget.
+        <emphasis>budgeted</emphasis> balance sheet is based on the predictions made in the budget.  For more details, see <ulink url="&url-docs-C;manual/report-classes.html#report-budget"><quote>Budget Reports</quote> in the manual</ulink>
```
fellen, line 209: IMHO most of the `<sect1 id="budget_reporting1">` is still descriptive and belongs into the manual. Here I would expect rules to get the best results  from using the reports like e.g. `At first use report x to check for input errors.  Some common rules are …`

C/manual/ch_Reports.xml
```diff
@@ -359,8 +359,8 @@
         </listitem>
 
         <listitem>
-          <para>Maximum Slices: Controls the maximum number of slices displayed in a piechart - other accounts will
-            be placed in a slice marked Other.
+          <para>Maximum Slices: Controls the maximum number of slices displayed in a piechart &minus; other accounts
```
fellen: IMHO `… in a piechart. All other accounts …` is easier to understand.

C/manual/ch_Reports.xml
```diff
@@ -3778,7 +3778,179 @@ Income (type INCOME)
...
+            <glossdef>
+              <para>The sum of Existing Equity and New Equity
+              </para>
+            </glossdef>
+          </glossentry>
+
+          <glossentry>
+            <glossterm>Total Liabilities + Equity</glossterm>
+            <glossdef>
+              <para>Self Explanatory. This will equal Total Assets.
```
fellen, line 3949: … if the book is balanced.

C/manual/ch_Reports.xml
```diff
@@ -3843,81 +4015,83 @@ Income (type INCOME)
 
           <itemizedlist>
             <listitem>
-              <para><emphasis>Show budget</emphasis> - include the budgeted values for the period
+              <para><emphasis>Show budget</emphasis> &minus; include the budgeted values for the period
```
fellen: Do you think you can convert the itemzedlist into a glosslist and drop the `-`? It seems the original author did not know variablelist or glosslist and used it as separator between the term and its explanation.

fellen:
a few more thoughts:

- Are these reports sharing the same report options (as Budget Report)?
- If not, add a options sections to the new reports;
- If yes, should we move that report or at least its option section to the begin?
- As most reports dislay a table, should we distinguish table rows and columns?
- Should we add Screenshots? We can store an example data file for their generation in the data directory.

C/manual/ch_Reports.xml
jralls: > … if the book is balanced.

Always. The Balance Sheet Report gets a book into balance by inserting a calculated Equity account, Unrealized Gains (or Losses depending on whether Equity is less or greater than Assets - Liabilities) to balance an out-of-balance book. Everything else in the report exactly reflects the balances in the accounts at the end of the selected date.

C/manual/ch_Reports.xml
jralls: Much better to use the entity: It's both easier to enter (key combinations are localized and OS-specific) and especially in the case of `&mdash;` (U+2014), `&ndash;` (U+2013), `&minus;`, (U+2011), and `&hyphen;` (U+2010, non-breaking U+2011) and the ASCII hyphen-dash (U+002d) that are hard to tell apart in source strings.

But I think it's better still to avoid using any of them except `&minus;` in its arithmetic meaning because we want the documentation translated. Correct usage of the various dashes and hyphen in English can be so obscure that many native English speakers don't even know that there's a difference (it's wrong throughout this PR: `&minus;` is for arithmetic and shouldn't be used in a sentence). Expecting non-native English speakers to correctly understand and translate them is asking a lot. em-dashes are usually a substitute for parentheses. That's the case here, so `(see <xref linkend="report-common"/>)` would be better.

jralls:
Apropos hyphens and dashes, using any of them in the form of `Assets&pickyourdash;General` is wrong. A colon would be better, but the repetition is tiresome and the layout obscures that `Assets - General` is a section header and the rest are summary lines at the end of that section. Laying the thing out as a glossary makes it even worse. @datajunkie-eng's original submission was very much better.

C/manual/ch_Reports.xml
fellen: Interesting, @jralls . There seems to be regional differences. In german books parentheses are used in formulas, but rarely in text.

We should try to find some consent and update the wiki section, which was written by an Australian IIRC.

C/manual/ch_Reports.xml
jralls: @fellen, I think you're driving massive overuse of [`<glosslist>`](https://tdg.docbook.org/tdg/4.5/glosslist). As the docs say, "Using a [GlossList](https://tdg.docbook.org/tdg/4.5/glosslist) in running text, instead of a [VariableList](https://tdg.docbook.org/tdg/4.5/variablelist), for example, maintains the semantic distinction of a Glossary. This distinction may be necessary if you want to automatically point to the members of the list with [GlossTerm](https://tdg.docbook.org/tdg/4.5/glossterm)s in the body of the text."  A [glossary](https://www.merriam-webster.com/dictionary/glossary) is "a collection of textual [glosses](https://www.merriam-webster.com/dictionary/glosses) or of specialized terms with their meanings". That isn't even remotely related to what you're trying to use it for.

C/manual/ch_Reports.xml
jralls: Punctuation is very language specific. I wouldn't expect anything beyond full stops to transfer even between closely related languages like German, French, and English, never mind non-european ones. Even native speakers and writers of English have trouble with the finer points of punctuation, and there are lots of heated disputes about some practices, e.g. the [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma). As it happens another hot dispute is how to set off a parenthetical expression. Sorry, I couldn't find a good reference for this, so here goes: The basic form that a lot of people think should be the only one is to use a pair of commas. The other inline choices are parentheses and em-dashes. One can also use footnotes or endnotes, an especially good idea if the aside will be lengthy.

The section in the wiki page about dashes just points to https://www.thepunctuationguide.com/index.html. I have no idea who wrote it but the sections are explanatory rather than style guidance. The flavor of English used by the original author doesn't seem important at least in that case.

datajunkie-eng: @jralls and @fellen I'm fine continuing however you want with this bit of it, but I should mention that the changes outside the Budget Balance Sheet were inadvertent. I'd forgotten to limit the scope of my search/replace.  Perhaps I should undo all these other changes?  I'm not sure I want to be on the hook for this entire xml file at this time.

C/manual/ch_Reports.xml
fellen: Well, this is my first review of a report. While in dialogs glosslists make sense because often one control depends on others. in reports the [`<variablelist>`](https://tdg.docbook.org/tdg/4.5/variablelist-x) might be sufficient.

jralls: ["variablelist — A list in which each entry is composed of a set of one or more terms and an associated description"](https://tdg.docbook.org/tdg/4.5/variablelist.html) (Heh, complete with a misused emdash!)
That's no more applicable than `glosslist`, not to dialogs and not to reports. 

Plain old `itemizedlist` is generally the right choice, but only so far as a list is appropriate. E.g.
```
<section>
<title>Assets</title>
<para>The first section of the report lists the asset accounts, displayed hierarchically, with their balances projected to the end of the budget period. At the end of the section &app; will print the following totals:
<itemizedlist>
<listitem><title>Existing assets</title><para>Sum of all selected account balances at the beginning of the budget period. This is what a Balance Sheet Report would report as total assets for the same selected accounts on that date.</para>
<!-- … --> 
</itemizedlist>
</para>
</section>
```
While `glosslist` and `variablelist` assign the wrong semantic meaning in this case, even trying to wrap bits of documentation in semantic tags is a relic of the 1990s. The semantic web fantasy didn't pan out. Stop wasting everyone's time and just focus on getting the right words in place with a reasonable layout.

fellen: Nice idea, but
```
C/manual/index.docbook:3792: element listitem: validity error : Element listitem content does not follow the DTD, expecting (calloutlist | glosslist | bibliolist | itemizedlist | orderedlist | segmentedlist | simplelist | variablelist | caution | important | note | tip | warning | literallayout | programlisting | programlistingco | screen | screenco | screenshot | synopsis | cmdsynopsis | funcsynopsis | classsynopsis | fieldsynopsis | constructorsynopsis | destructorsynopsis | methodsynopsis | formalpara | para | simpara | address | blockquote | graphic | graphicco | mediaobject | mediaobjectco | informalequation | informalexample | informalfigure | informaltable | equation | example | figure | table | msgset | procedure | sidebar | qandaset | task | anchor | bridgehead | remark | highlights | abstract | authorblurb | epigraph | indexterm | beginpage)+, got (title para )
Document /home/frank/git/gnucash-docs-old/C/manual/index.docbook does not validate
```
Normal lists allow no title in their elements, only for the whole list.  That is the reason why we use lists with term and definition part. I assume we still want get rid of constructs like
 `<emphasis role="bold">Term</emphasis>, a changing combination of colons or dashs and spaces; Definition`

fellen:
> Apropos hyphens and dashes, using any of them in the form of `Assets&pickyourdash;General` is wrong. A colon would be better, but the repetition is tiresome and the layout obscures that `Assets - General` is a section header and the rest are summary lines at the end of that section. Laying the thing out as a glossary makes it even worse. @datajunkie-eng's original submission was very much better.

Right, the list should be split in groups Assets, Liabilities, Equity
and the group name should be dropped from their elements terms.
Perhaps the 3 lists should be wrapped into another list.

jralls:
Merged, on the grounds that it's an improvement in spite of the nits. The file suffixes changed to docbook in the interim so Github can't see the merge.

PR #331, opened 2023-10-02
Redefine transaction

flywire:
Describe what a transaction actually is. Informed by https://www.investopedia.com/terms/t/transaction.asp .

If `key-value frame` is required it should be written in simpler terms.

### DIFF
--- a/C/guide/gnc-glossary.xml
+++ b/C/guide/gnc-glossary.xml
@@ -452,8 +452,9 @@
   <glossentry id="gnc-gl_transaction">
     <glossterm>Transaction</glossterm>
     <glossdef>
-      <para>A transaction consists of a date, a description, an ID number, a list of two or more splits, and a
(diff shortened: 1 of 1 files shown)


fellen:
Hm, perhaps we should distinguish like [wikipedia](https://en.wikipedia.org/wiki/Transaction):
'''Transaction''' or '''transactional''' may refer to:
==Commerce==
*[[Financial transaction]], an agreement, communication, or movement carried out between a buyer and a seller to exchange an asset for payment
*Debits and credits in a [[Double-entry bookkeeping system]]
…
@jralls ?

jralls:
This PR is misguided. The definition of transaction here is how it's used in GnuCash, which is appropriate for GnuCash's glossary. Users should consult a [dictionary](https://www.oed.com/search/dictionary/?scope=Entries&q=transaction) for the plain-language meaning or an [accounting text](https://www.investopedia.com/terms/t/transaction.asp) for the accounting meaning.

PR #334, opened 2024-02-06
prepare Docbook XSL-Stylesheets

CWehli:
Inspired by the ideas of TANIGUCHI Yasuaki, these adjustments were made to the existing XSL stylesheets:

* reorganize the html-targets
* do pdf-spezific customizing for
  * lists
  * titelpage
  * paperformat depending LANG
* optimize admonitons
  * icons
  * spacing
  * font-size

For these commits I used parts from PRs #307, #308, #309, #310 and #311 as templates.


fellen:
> * rename the customized stylesheets to gnc-custom-*

Why? The whole level are our customizations – borowed decades ago from `GNOME Documentation Project`
(gdp 1[?]) That evolved to gdp 2 (in between abandoned) and yelp-xsl (used by yelp).
Only the subdirectory 1.79.2 is from docbook.

CWehli:
> Why? The whole level are our customizations

I tried to get a sorting in the customization level. This is only cosmetic and can be omitted.

xuxinhang:
Maybe using the path like `html-markup/glossary.xsl` instead of the file name prefix like `html-glossary.xsl` could be clearer, because put something like `fo-glossary.xsl` (maybe someday) together seems a mess.

Or even put all `html-*.xsl` into one single file like `markup-html-custom.xsl`? Each of `html-*.xsl` seems too short and fragmented.

jralls:
@CWehli is this still something you mean to get back to?

PR #328, opened 2023-09-02
Guide(zh): Simplified Chinese translation for the Guide (Stage 2)

xuxinhang:
Continue from PR #323 .

This PR contains translated texts for GnuCash Concept Guide, from English version to Simplified Chinese.

PR #336, opened 2024-04-15
Guide(zh): SC translation for the Guide, Chapter Investments.

xuxinhang:
Further Simplified Chinese translation to PR #328.

Newly-added translation contains Chapter Investment in Concept Guide, with some items in other chapters updated accordingly.

### DIFF
--- a/zh/guide/ch_invest.docbook
+++ b/zh/guide/ch_invest.docbook
@@ -13,207 +13,171 @@
     Updated	Dave Herman <dvherman@bigfoot.com>
   Originally designed by Carol Champagne.
   Translators:
+    Xinhang Xu <xuxinhang4567@outlook.com>
     (translators put your name and email here)
 -->
 <chapter id="chapter_invest">
-  <title>Investments</title>
+  <title>投资</title>
 
-  <para>This chapter explains how to manage your investments with &app;. Most people have an investment
-    plan, whether its just putting money into a certificate of deposit (<acronym>CD</acronym> for
-    short) account, investing through a company sponsored plan at your workplace or buying and
-    selling stocks and bonds through a brokerage. &app; gives you tools to help you manage these
-    investments such as the <emphasis>Price Database</emphasis> which allows you to record changes
-    in the prices of stocks you own.
+  <para>本章介绍怎样使用&app;管理投资。大部分用户都有投资计划，不论是银行存单、就职企业的激励计划还是通过经纪人买卖的股票和债券。&app;会提供给你有助于管理这些投资的工具，例如<emphasis>价格数据库</emphasis>就可以帮助你记录所持股票的价格起落。
   </para>
 
   <sect1 id="invest_concepts1">
-    <title>Basic Concepts</title>
+    <title>基本概念</title>
 
-    <para>An investment is something that you purchase in the hopes of generating income, or that you hope to
-      sell in the future for more than you paid. Using this simple definition, many things could be
-      considered investments: the house you live in, a valuable painting, stocks in publicly traded
-      companies, your savings account at the bank, or a certificate of deposit. These many types of
-      investments will be discussed in this chapter in terms of how to track them using &app;.
+    <para>投资是你希望用于生产收入或希望未来以更高价格卖出而买入的东西。根据这个简单的定义，许多东西均可被视作投资：你居住的房屋、贵重的画作、上市公司的股票、储蓄存款账户或银行存单。本章会讨论怎样使用&app;跟踪这些纷杂多样的投资类型。
     </para>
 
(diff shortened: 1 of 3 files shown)


zh/guide/ch_invest.docbook
```diff
@@ -343,165 +266,136 @@
...
 
-      <para>The following is a somewhat more complicated example of setting up &app; to track your investments,
-        which has the advantage that it groups each different investment under the brokerage that
-        deals with the investments. This way it is easier to compare the statements you get from
-        your brokerage with the accounts you have in &app; and spot where &app; differs from the
-        statement.
+      <para>下面是一个稍显复杂的示例，展示了怎样设置用于跟踪投资的&app;的科目结构。其优势在于将每种不同的投资对象分组在处理投资交易的经纪商之下。这样更便于比对经纪商给出的账单与&app;中已有的科目交易记录，并定位&app;与账单间的差异。
+      </para>
+        
+      <simpara>
```
fellen: This unintended insertion is breaking the check.
`ninja zh-guide-check` returns
`index.docbook:305: element simpara: validity error : Element screen is not declared in simpara list of possible children`

Because we are using `<xi:include href="…" />` check returns an unusable location.
To find the right  location I have run
```
grep -inr '<screen>' *.docbook
grep -inr '<simpara' ch_invest.docbook
```

zh/guide/ch_invest.docbook
```diff
@@ -343,165 +266,136 @@
...
+                免税
+ 费用
+    投资费用
+       手续费
+          海通证券
+          天天基金
+       管理费用
+          海通证券
+          天天基金</screen>
+      </simpara>
```
fellen: and the corresponding closing tag needs removement, too.

fellen (changes requested):
There are a few syntactical errors.
To avoid that you should run after each change in zh/guide `ninja zh-guide-check`

I explained inline how I found the first one and will leave the others for you.
We don't want to encapsule screens in any paragraphs.

fellen (approved):

zh/guide/ch_invest.docbook
```diff
@@ -13,327 +13,250 @@
...
       <tip>
-        <para>You can also run the <guilabel>New Account Hierarchy Setup</guilabel> assistant by creating a new
-          &app; file.
+        <para>你也可以通过创建新的&app;文件运行<guilabel>新建科目层次结构</guilabel>向导。
         </para>
       </tip>
 
       <figure>
-        <title>A New File Containinng Only The Default Investment Accounts</title>
+        <title>只包含默认投资科目的新文件</title>
```
fellen, line 259: Are you planning to create this screenshots in zh_CN?
See https://wiki.gnucash.org/wiki/Documentation_Improvement#Screenshots_And_Other_Images

fellen (commented):
I was  able to create it for yelp and left a note about screenshots inline. Do you want to generate the screenshots now or in a later PR?

One issue I watched seems not caused by your changes. I report it here just in case you might have an idea, what caused it.
Running `ninja html` I got
```
[2/7] Generating htmltrigger
No template for "/book/bookinfo" (or any of its leaves) exists in the context named "title" in the "zh" localization.
No "zh" localization of "History" exists; using "en".
No "zh" localization of "Title" exists; using "en".
No "zh" localization of "Date" exists; using "en".
```
I am not sure if our stylesheets are outdated or where we are misssing that translations.

zh/guide/ch_invest.docbook
xuxinhang: Yes. Maybe it's better to create screenshots later. Because there's still some work to choose better translation (trade-off between the literal and free translation) for terms (for example how commercial software or textbooks widely used by Chinese accountants express those concepts). GnuCash's expression seems to be organized for US accounting, somewhat. It seems not so interesting to take screenshots each time wanting to change a term, let alone it's even more inconvenience to update GUI translation.

fellen: I know that problem, too. E.g. English is missing accounting terms, which are common in all other European languages.

And yes most parts of the guide were written by English natives. That results in missing some aspects.
While @CWehli is translating the manual to German, we often have to improve the english version to cover other views.

In general out target is to move from [US-GAAP](https://en.wikipedia.org/wiki/Generally_Accepted_Accounting_Principles_(United_States)) to [IFRS](https://en.wikipedia.org/wiki/International_Financial_Reporting_Standards).

So if the english guide does not cover important aspects, feel free to send PRs.

fellen:
Merged, thanks!

FYI my feeling says instead of the over 20 years old xsl/l10n.xml our build system should use the files in xsl/1.79.2/common,but that is a separate task.

PR #338, opened 2024-05-01
Revert commit 180624f and use entity url-wiki

CWehli:


docbook/gnc-gui-C.dtd
```diff
@@ -448,9 +448,9 @@
 
 <!ENTITY % untranslated 'Dear Reader, we are really sorry, but nobody offered a translation
   for the following parts. If you would like to help us, follow the instructions at <ulink
-    url="&url-wiki-L10N-docs;">
-    <citetitle>&lbl-wiki; &lbl-wiki-L10N-docs;</citetitle></ulink>.
-  Questions can be discussed at the <<email>gnucash-XX@gnucash.org</email> mailing list.'>
+    url="https://wiki.gnucash.org/wiki/Translation#Translating_the_GnuCash_Guide_and_Help">
```
fellen: The main reason for my commit was to use the direct link to
`https://wiki.gnucash.org/wiki/Documentation_Translation `instead of the indirect link over
`https://wiki.gnucash.org/wiki/Translation#Translating_the_GnuCash_Guide_and_Help`, which again should get renamed to `GnuCash_Guide_and_Manual`.

fellen: Also include after
```
<!--
  Some global declarations which are used in the files. For easier
  maintenance they are collected here.
```
a reminder
`
  Note that this entities are not available in gnc-gui-*.dtd files
`

fellen: > which again should get renamed to `GnuCash_Guide_and_Manual`.

Sorry, that was a note for the wiki,  not the docs. [Edit:] And `GnuCash_Manual_and_Guide` seems better there.

fellen:
Danke!

Wir sollten aber mal im Auge behalten, warum yelp die URL statt des Labels anzeigt, genauso wie die fehlenden `<footnoteref>`s.

PR #342, opened 2024-09-22
Security Assistant Field Order

flywire:
Describe fields in same order as screen

PR #343, opened 2024-09-23
Security Assistant Field Order

flywire:
Describe fields in same order as form

Replaces #342

### DIFF
--- a/C/manual/ch_Tools_Assistants.docbook
+++ b/C/manual/ch_Tools_Assistants.docbook
@@ -5458,6 +5458,21 @@
 
     <itemizedlist>
       <listitem>
+        <para><emphasis role="bold">Name:</emphasis> The full name of the commodity is a recognizable name such as
+          <emphasis>US Dollars</emphasis> or <emphasis>IBM Common Stock</emphasis>.
+        </para>
+      </listitem>
+
+      <listitem>
+        <para><emphasis role="bold">Symbol:</emphasis> Indicates the symbol or abbreviation for the commodity.
+          This is usually the ticker symbol (for stocks) or other unique abbreviation for the
+          commodity. If the commodity is traded on any public exchange, it is important to use the
+          same identifier used on that exchange. For national currencies the symbol is the ISO-4217
+          currency code.
+        </para>
+      </listitem>
+
+      <listitem>
         <para><emphasis role="bold">Type:</emphasis> Categories for organizing securities. &app; has the following
           built in:
           <itemizedlist>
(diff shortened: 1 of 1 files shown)


C/manual/ch_Tools_Assistants.docbook
```diff
@@ -5457,6 +5457,21 @@
     </para>
 
     <itemizedlist>
+      <listitem>
+        <para><emphasis role="bold">Name:</emphasis> The full name of the commodity is a recognizable name such as
```
fellen, line 5461: `<emphasis role="bold">Full name</emphasis>`

C/manual/ch_Tools_Assistants.docbook
```diff
@@ -5457,6 +5457,21 @@
...
+          <emphasis>US Dollars</emphasis> or <emphasis>IBM Common Stock</emphasis>.
+        </para>
+      </listitem>
+
+      <listitem>
+        <para><emphasis role="bold">Symbol:</emphasis> Indicates the symbol or abbreviation for the commodity.
```
fellen, line 5467: `<emphasis role="bold">Symbol/abbreviation</emphasis>`

fellen (commented):
Not only the order changed:

- This labels were also renamed to remove ambiguity.

- Also the colons were removed from the labels — in the whole dialog.
- I wonder, if you could also insert a screenshot like in the [german version](https://www.gnucash.org/docs/v5/de/gnucash-manual/tool-security-edit.html#tool-commodity)
- and perhaps apply the new style with `<gloss*>` instead of  `<emphasis role="bold">`?

flywire:
> Not only the order changed:
 
But it's complicated and I suppose ultimately this PR will be closed without merging.

> * This labels were also renamed to remove ambiguity.

How about a critique?
* `Display` would be clearer than `Display symbol` to remove confusion with `Symbol/abbreviation`, the tooltip has the details (which btw doesn't refer to Symbol/abbreviation).
* `Namespace` means `Namespace` apparently from dev response on mail list. It's not in any English dictionary and after a web search it won't be any clearer. `Type` (or even category) was very clear. I'm not sure what is built-in  either.

> * Also the colons were removed from the labels — in the whole dialog.

Doesn't seem relevant to this section which has a consistent style with the rest of the document.

> * I wonder, if you could also insert a screenshot like in the [german version](https://www.gnucash.org/docs/v5/de/gnucash-manual/tool-security-edit.html#tool-commodity)

Screenshots from random version/OSs look scrappy ([English version](https://www.gnucash.org/docs/v5/C/gnucash-manual/tool-security-edit.html)).

> * and perhaps apply the new style with `<gloss*>` instead of  `<emphasis role="bold">`?

Isn't this something that could be scripted in the repo across all docs/languages? This is one of the downsides of complicated markup compared to something like markdown.

fellen:
@CWehli missed I your backport of ch_Tools_Assistants.docbook or have you aready something prepared?

fellen:
> How about a critique?
> …

We try to avoid renaming english terms because over 60 translators of the program and the documenters have to apply the change, too.

fellen:
> * `Namespace`

Right, Google translate would understand `name space`, but not `namespace`. @jralls perhaps to fix after the string freeze is lifted?

jralls:
> Right, Google translate would understand name space, but not namespace. @jralls perhaps to fix after the string freeze is lifted?

Well, Google translate would understand "name" and "space" but that wouldn't necessarily convey the same meaning. After all "namespace" is computer science jargon so it's not going to be meaningful to normal people anyway.  I think "type" is a bit too restrictive, how about "Commodity Group"?

jralls:
Thanks!

PR #345, opened 2024-10-17
Update Appendix A for finance-quote 1.63

pkryger:
A few weeks ago @bpschuck [asked for help](https://github.com/finance-quote/finance-quote/discussions/423) with updating `Gnucash` for recent version of `finance-quote`. This is a complitentary PR to https://github.com/Gnucash/gnucash/pull/2033.

I decided to move modules that are not a part of the list into a new section for deprecated sources. IIUC this is still usefull information for a case when an old `.gnucash` file (or a database) has been configured with one of the deprecated sources.

PR #346, opened 2024-10-23
Remove Finance::Quote appendix

pkryger:
A few weeks ago @bpschuck [asked for help](https://github.com/finance-quote/finance-quote/discussions/423) with updating `Gnucash` for recent version of `finance-quote`. This is a complimentary PR to https://github.com/Gnucash/gnucash/pull/2033.

Note this is a follow up PR to #345. During review of the original one it had been decided to move the content to Wiki. Please see https://wiki.gnucash.org/wiki/Online_Quotes#Finance::Quote_Sources for the removed content (with updates as in the original PR).

I updated references to the moved content with Wiki counterparts. I have not removed not used definitions from DTD as they are still used by a few translations. I think when translation catch up these can be removed.


jralls:
Thanks!

PR #347, opened 2024-11-19
Guide(zh):  Text translation for chapter currency & Screenshot translation for chapters before invest. 

xuxinhang:
(1) Translation for chapter currency.

(2) Screenshot translation for chapter txn, basics, accts, loan, capgain, credit-card and checkbook. All images are taken under DPI of ~200%, but have been compressed for less file sizes, using pngquant.

(3) I added some XSLT templates to make screenshot images scale on proportion to the physical width displayed on the actual screen, and not exceed the assigned upper-limit width simultaneously. The code is language-specific.
 
(4) During screenshot translation, I updated the po file for more habitual and accurate expressions. May I make PR to gnucash main repo, instead of using WebTranslator, for po file updating?


jralls:
Thanks!

PR #349, opened 2025-01-22
manual:ch_Transactions: Define entities for colors used in the Import Matcher dialog box

CWehli:
Today, the color of table rows is set by "processing instructions". By using entities instead, it is easier to understand how the rows are colored.

### DIFF
--- a/C/manual/ch_Transactions.docbook
+++ b/C/manual/ch_Transactions.docbook
@@ -2496,32 +2496,25 @@ Translators:
                   </entry>
 
                   <entry>
-<?dbhtml bgcolor="#FFD700"?>
-<?dbfo bgcolor="#FFD700"?>
-                    Gold
+                    &color-intervention-probably-required; Gold
                   </entry>
 
                   <entry>
-<?dbhtml bgcolor="#FFD700"?>
-<?dbfo bgcolor="#FFD700"?>
-                    No matches, flagged for input and no transfer account matched &ndash; Import as
-                    a new transaction. Intervention required to select a transfer account.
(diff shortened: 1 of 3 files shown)


CWehli:
I wonder if we should use named colors from this site: [W3C School](https://www.w3schools.com/colors/colors_names.asp)
brown: #A52A2A
gold: #FFD700
darkseagreen: #8FBC8F
cornflowerblue: #6495ED

jralls:
Thanks. 

One thing though: The intervention-required color looks a lot more Red than Brown to me, notwithstanding the X11 name for it. What do you think? @fellen, what about you?

CWehli:
@jralls 
I agree with you, but brown1 is mentioned in gnucash-fallback.css and that's why I chose this color.

> One thing though: The intervention-required color looks a lot more Red than Brown to me, notwithstanding the X11 name for it. What do you think?

jralls:
@CWehli, of course it is. The CSS color names are derived from the X11 ones and technically it is a brown because it has some blue turned on. But color arcana is a bit lost on users. It looks red (except of course to the 10% of men who are red-green colorblind) and that's obviously intentional: Westerners are used to red meaning error.

> I wonder if we should use named colors from this site: [W3C School](https://www.w3schools.com/colors/colors_names.asp)
> brown: #A52A2A

That would be the wrong direction, it looks actually brown. We could change to orange-red #FF4500, crimson #DC143C, or red #FF0000 if that would make you happier with calling it red. But the technical standard color name, whether it's X11 or W3C, doesn't really matter that much.

PR #350, opened 2025-03-25
L10N:de ch_import_business_data.docbook translation corrected

warrel040:
- In the CSV-File instead of "J" one has to use "Y" for "YES" in Germany.
- The name of the tax table must be used instead of the tax rate.


jralls:
Thanks!

PR #355, opened 2025-07-05
cmake: Since the variable ${PACKAGE} only exists once and does not appear to be needed, we can drop it.

CWehli:


jralls:
No, it's supposed to be `PROJECT_NAME`, so that the path is `share/doc/gnucash-docs`, matching the installation directory. I've fixed it.

PR #357, opened 2025-11-04
Multi-split.csv Transfer_Account field

flywire:
Should be Transfer Account field not Account field.

### DIFF
--- a/C/manual/ch_Transactions.docbook
+++ b/C/manual/ch_Transactions.docbook
@@ -1848,7 +1848,7 @@ Translators:
                       <term>split specific fields</term>
 
                       <listitem>
-                        <para>account and deposit/withdrawal as bare minimum.
+                        <para>account (may set base account), transfer account and deposit/withdrawal as bare minimum.
                         </para>
                       </listitem>
                     </varlistentry>
--- a/C/manual/ch_Transactions.docbook
+++ b/C/manual/ch_Transactions.docbook
@@ -1790,9 +1790,9 @@ Translators:
 
               <listitem>
                 <para><emphasis role="bold"><guilabel>Account</guilabel></emphasis>: If the data you are importing does
-                  not specify the source account for the transaction an account can be selected from
-                  the drop down list will import all transactions to the selected account. Leave
-                  blank if your imported data specifies a source account;
+                  not specify the source account for each transaction a base account can be selected
+                  from the drop down list. All transactions will then be imported into this account.
+                  Leave blank if your imported data specifies a source account;
                 </para>
               </listitem>
 
--- a/C/manual/ch_Transactions.docbook
+++ b/C/manual/ch_Transactions.docbook
@@ -1864,7 +1864,7 @@ Translators:
 
                 <example>
                   <title>Sample Multi-split.csv</title>
-<programlisting language="csv">Date,Description,Deposit,Account
+<programlisting language="csv">Date,Description,Deposit,Transfer_Account
 05/03/2006,Grocery Store,-45.21,Assets:Checking
 ,,45.21,Expenses:Groceries
 14/03/2006,Employers R Us,670.00,Assets:Checking


C/manual/ch_Transactions.docbook
```diff
@@ -1848,7 +1848,7 @@ Translators:
                       <term>split specific fields</term>
 
                       <listitem>
-                        <para>account and deposit/withdrawal as bare minimum.
+                        <para>account (may set base account), transfer account and deposit/withdrawal as bare minimum.
```
gjanssens, line 1851: The transfer account is optional. If not set, gnucash will attempt to guess it based on already existing transactions with or without bayesian heuristics. An account (or the base account) on the other hand is mandatory.

C/manual/ch_Transactions.docbook
```diff
@@ -1864,7 +1864,7 @@ Translators:
 
                 <example>
                   <title>Sample Multi-split.csv</title>
-<programlisting language="csv">Date,Description,Deposit,Account
+<programlisting language="csv">Date,Description,Deposit,Transfer_Account
```
gjanssens, line 1867: In multi-split mode, Transfer Account is not a selectable column type. You can only select Account from the list.

gjanssens (changes requested):

C/manual/ch_Transactions.docbook
flywire: That's logical because there is no transfer account at the split level. Thanks.

flywire:
See https://lists.gnucash.org/pipermail/gnucash-user/2025-November/118164.html

jralls:
@flywire, linking to a mailing list thread doesn't in any way resolve @gjanssens review comments. Are you going to fix them?

flywire:
I will address this, but it requires further consideration and some rework next year.

My current understanding is that, by default, the CSV import module operates with a single-entry accounting file, while the Multi-split option is intended for a double-entry accounting file. Although a double-entry file supports splits, the emphasis should remain on double-entry accounting rather than on the split structure itself.

I need to examine the handling of single-entry files and splits in more detail before finalising this.

---

The term Multi-split is used minimally within the software and documentation, but the community appears to recognise it, albeit with varying interpretations.

Another aspect worth noting is that, although a simple double-entry transaction involving two accounts is not formally considered a split (each side posts to a single account), each account entry is still represented internally as a split in the code. This appears to align with the understanding of some community members.

jralls:
@flywire, if you're getting tripped up over multi-split, think of it in terms of the register: Basic and Transaction Journal. A single line per transaction file can import only two splits per transaction: One to the Account and the other to the Transfer Account. If there's no Account column GnuCash allows you to set the Account field in the assistant page; if there's no Transfer Account column GnuCash will use the Generic Import Matcher to let the user record the account for that split and if they don't then the other split is to the imbalance account.
Multiple-lines-per-transaction CSV files have a line for each split and there must be one account column containing the account for that split. There's no Transfer Account column in that case because GnuCash isn't creating any splits.

flywire:
@gjanssens and others. Can you check and comment on logic, technical and/or style? These changes would need to go into the CSV Importer and be reflected in the docs.

---

**Overview**

The GnuCash CSV Importer UI and docs are confusing because the terminology is not clear and the Account field (for single-line mode) appears before the Multi-split checkbox decision point that determines the import mode.

Single-line mode (one line = one transaction, two accounts):
* Source Account: Set once in the assistant or via CSV column
* Transfer Account: CSV column or Import Matcher (defaults to Imbalance)

Multi-line mode (one line = one split, 2+ accounts):
* All data per-line; no common Account field
* Groups consecutive lines by identical/blank transaction info (Date, Description, etc.)

Every transaction needs: Date, Description, Amount(s), Account(s).

Single-line/Multi-line terms are equivalent to Simple/Split Transaction terms used in the Guide for 2/2+ accounts respectively and also related to Register View styles for Basic Ledger and Transaction Journal.

**References:**
* https://github.com/Gnucash/gnucash/blob/stable/gnucash/gtkbuilder/assistant-csv-trans-import.glade#L46-L55
* https://github.com/Gnucash/gnucash/blob/stable/gnucash/gtkbuilder/assistant-csv-trans-import.glade#L527-L530

---

**Proposed UI changes:**
1. Rename Multi-split to Multi-line
2. Move Account field under Multi-line checkbox  
3. Dim/disable Account field when Multi-line enabled
4. Update tooltip-text
5. Update documentation to describe

**Multi-line tooltip-text:**

* Disable: Each CSV line is one complete transaction.
* Enable: Multiple consecutive lines form one transaction. Each line provides details for a single account or split, and the first line must include transaction‑specific columns (Date and Description as minimum).<br>
The importer groups consecutive lines into the same transaction until it finds a line where any transaction‑specific column (Transaction ID, Date, Number, Description, Notes, Transaction Commodity, or Void Reason) contains different, non‑empty information.

**Account tooltip-text:**

Apply a single source account to all imported transactions. The other side of the transaction can be provided in a Transfer Account column, or left for the Import Matcher to assign.

PR #360, opened 2026-07-05
Document CAGR columns in Investment Lots report

mcbridebt:
https://github.com/Gnucash/gnucash/pull/1956 added new CAGR columns to the Investment Lots report. 

This PR updates the documentation accordingly.


C/manual/ch_Reports.docbook
jralls: Your images except for the stacked bar chart are too big to print into the PDFs without scaling, so please add `width="160mm" height="450mm" align="center"` to all of the `<imageobject role="fo"><imagedata>` elements. Note that you need to add those on some figures.

C/manual/figures/Report_Investment_Lots.png
jralls, line 1: 2510x1149 wastes a lot of space and just gets resampled down--and if it doesn't it slows page loading. I don't think much bigger than 1024x768 is useful.

C/manual/ch_Reports.docbook
mcbridebt: fixed

C/manual/figures/Report_Investment_Lots.png
mcbridebt: fixed

C/guide/ch_reports.docbook
```diff
@@ -331,17 +331,18 @@
...
                 <imagedata fileref="figures/Report_Investment_Lots.png"
-                            srccredit="Brent McBride" width="&img-w;"/>
+                            srccredit="Brent McBride" width="1024"/>
               </imageobject>
 
               <imageobject role="fo">
                 <imagedata fileref="figures/Report_Investment_Lots.png"
-                            srccredit="Brent McBride"/>
+                            srccredit="Brent McBride"
+                            width="160mm" depth="450mm" align="center"/>
```
jralls: `height` was right, why did you change it?

mcbridebt: DocBook's `<imagedata>` uses
`depth` (not `height`) for the vertical dimension, and our FO stylesheet
reflects that. In `xsl/1.79.2/fo/graphics.xsl`, the `image.height` template only
reads `@depth` — there's no reference to `@height` anywhere in it. So
`height="450mm"` would be silently ignored (falling through to `auto`), whereas
`depth="450mm"` is what actually constrains the image height in the PDF.

So I think `depth` is what we want here to get the sizing you originally asked
for. Am I missing something — is there a reason `height` is preferred in our
setup? Happy to switch it back if so.

jralls: Well, when it's depth the images don't draw, when it's height they do. But maybe that's because it's being ignored and there's something wrong with "450mm".

mcbridebt: I looked into this with a local FO build (xsltproc) and I think we were both
half right. Our stylesheet's `image.height` template
(xsl/1.79.2/fo/graphics.xsl) only reads `@depth` — it never looks at `@height`.
I confirmed this by comparing the generated .fo:

- `depth="450mm"` → `<fo:external-graphic ... height="450mm">`
- `height="450mm"` → `<fo:external-graphic ... height="auto">` (value ignored)

So `height` "works" only because the 450mm is silently dropped and the
`width="160mm"` does all the constraining. With `depth`, the 450mm is honored —
and since it's larger than the 297mm A4 page height, FOP can't place the
viewport and drops the image. Your hunch about "something wrong with 450mm" was
right on the money.

Since the screenshot is landscape, width is the binding dimension anyway, so
I've dropped the vertical constraint and left `width="160mm" align="center"` on
both figures. That emits `height="auto"` with content scaled-to-fit.

One caveat: I could only verify the generated .fo geometry, not the final
rendered PDF — I don't have fop installed here. If you want a vertical cap, it
would need to be `depth` (not `height`) with a value that fits the printable
body height (roughly 246mm on A4 before the figure caption), but I haven't
render-tested a specific value, so I left it off to be safe.

jralls: `depth="150mm"` does indeed work but so does the default `auto` so let's go with that.

jralls:
Thanks!

PR #361, opened 2026-07-13
bug 799744 - Import transaction from CSV skips fields if CSV file has…

tjteixeira:
… backslash just before final quote in a quoted field

Add description of the new check box "Enable escape character" in section "6.16.5 Import CSV Transations" which is selected by default but if not set, escape characters (backslashes) will have no special meaning. In particular, backslashes immediately before a quote are included literally and do not prevent recognition of the quote at the end of the string.

Note that running xmlformat.pl changed line breaks on otherwise unmodified text.

Updated Figure 6.5 "The CSV Import Preview Pane" to show the new checkbox in the enabled state.


C/manual/figures/CSV_Import_transactions_Preview.png
jralls, line 1: Please crop out the checkerboard background.

C/manual/ch_Transactions.docbook
```diff
@@ -1800,15 +1800,23 @@ Translators:
...
+                        required characters into the text box and selecting the
+                        <guilabel>Custom</guilabel> checkbox. This may be used in combination with
+                        any of the predefined separators; If the <guilabel>Enable escape
+                        character</guilabel> checkbox is selected, when the backslash appears inside
+                        a quoted string, the next character is processed specially: backslash-quote
+                        becomes a single quote character; backslash-n becomes a newline characters,
+                        and backslash-backslash becomes a single backslash. If the <guilabel>Enable
+                        escape character</guilabel> checkbox is not selected, the backslash
+                        character is not treated specially. The <guilabel>Enable escape
+                        character</guilabel> checkbox is selected by default.
```
jralls: "Enable Escape Character" deserves a separate paragraph.

jralls (changes requested):

C/manual/ch_Transactions.docbook
tjteixeira: Added a paragraph break and pushed a change, amending the previous commit.

C/manual/figures/CSV_Import_transactions_Preview.png
tjteixeira: Cropped to leave out the border. Committed amending the previous commit and pushed.

jralls:
Thanks!

PR #363, opened 2026-08-06
Changed 'account ' to 'account type'

pgfaller:
These lists are actually account types.


jralls:
Thanks!

flywire:
> These lists are actually account types.

Excellent! This needs to be carried through to sorting in the Transaction report.

PR #364, opened 2026-09-12
Deleted the note about .xac files

pgfaller:
The original version seemed to be missing why the .xac files might appear.


jralls:
Well, it's implied in the follow-on, "if you upgrade from 2.2 to 2.4...", but 2.4 was released 10 Dec. 2010 so this note seems a bit obsolete. I suppose it could say "If you've upgraded from 2.2 or earlier and still have 20 year old backups around they'll have the &lt;.code&gt;xac&lt;/code&gt; extension instead of the &lt;.code&gt;gnucash&lt;/code&gt one on more recent backups." but I think it makes more sense to delete it.

pgfaller:
Deleted the (obsolete) note about .xac files.

jralls:
Please squash out the first commit, the second supersedes it. You don't need a `Signed-off` trailer unless you used a generative AI to help you write it. That can't be the case here, you're just removing a paragraph.

jralls:
Thanks!

PR #365, opened 2026-09-15
Clarify QIF date formats and ISO YMD fixed in 5.17

brainbuz:
This is the documentation change for: https://github.com/Gnucash/gnucash/pull/2308

It clarifies which date formats in QIF files are supported, and after the fix in pull 2308 4 digit years in the ISO format are unambiguous


jralls:
https://github.com/Gnucash/gnucash/pull/2308 was rejected because it's the wrong way to fix the bug.

